PR #1, opened 2021-02-22
clean up cum and res plots, add better data source for examples plots

CompRhys:
1. passing y_true and y_pred to `cum_res` and `cum_err`
2. remove `cum_res_cum_err`
3. add better data source
4. split out the approximate density into a new function -> if small amounts of data we can use gaussian KDE instead


janosh:
Good stuff! Thanks @CompRhys 👍

Issue #17, opened 2022-02-12
Literal (typing) import error - incompatibility with Google Colab (Python 3.7)


sgbaird:
[![Open In Colab](https://colab.research.google.com/assets/colab-badge.svg)](https://colab.research.google.com/drive/1hnyQ-6D3TjpS6stxR86hMhtmCzsBYCQJ?usp=sharing)

```python
!pip install ml-matrics
from ml_matrics.parity import density_hexbins
```

> ```python
> ---------------------------------------------------------------------------
> ImportError                               Traceback (most recent call last)
> [<ipython-input-1-48d4b62e713f>](https://localhost:8080/#) in <module>()
> ----> 1 from ml_matrics.parity import density_hexbins
> 
> 1 frames
> [/usr/local/lib/python3.7/dist-packages/ml_matrics/__init__.py](https://localhost:8080/#) in <module>()
>       1 from .correlation import marchenko_pastur, marchenko_pastur_pdf
>       2 from .cumulative import add_dropdown, cum_err, cum_res
> ----> 3 from .elements import (
>       4     count_elements,
>       5     hist_elemental_prevalence,
> 
> [/usr/local/lib/python3.7/dist-packages/ml_matrics/elements.py](https://localhost:8080/#) in <module>()
>       1 from __future__ import annotations
>       2 
> ----> 3 from typing import TYPE_CHECKING, Any, Literal, Sequence
>       4 
>       5 import matplotlib.pyplot as plt
> 
> ImportError: cannot import name 'Literal' from 'typing' (/usr/lib/python3.7/typing.py)
> 
> ---------------------------------------------------------------------------
> NOTE: If your import is failing due to a missing package, you can
> manually install dependencies using either !pip or !apt.
> 
> To view examples of installing some common dependencies, click the
> "Open Examples" button below.
> ---------------------------------------------------------------------------
> ```
>

sgbaird:
Simple fix suggested [here](https://stackoverflow.com/a/67193166/13697228).

> ## Using `Literal` in all Python versions (1)
> 
> `Literal` was added to `typing.py` in 3.8, but you can use `Literal` in older versions anyway.
> 
> First install `typing_extensions` (`pip install typing_extensions`) and then
> ```
> from typing_extensions import Literal
> ```
> 
> This approach is supposed to work also in Python 3.8 and later.
> 
> ## Using `Literal` in all Python versions (2)
> 
> For completeness, I'm also adding the try-except approach to import `Literal`:
> ```
[...]

PR #18, opened 2022-02-12
python-requires>=3.8

sgbaird:
#17 #15


janosh:
Good catch, thanks for fixing!

PR #20, opened 2022-02-14
Add `plot_structure_2d()` in new module `ml_matrics/struct_vis.py`

janosh:
Adds `ml_matrics/struct_vis.py` with `ase`-inspired `plot_structure_2d()` but works natively with `pymatgen` `Structure`s, `matplotlib` being the only other dependency.

**Example**

```py
from pymatgen.ext.matproj import MPRester
from ml_matrics.struct_vis import plot_structure_2d

struct = MPRester().get_structure_by_material_id('mp-12712')
plot_structure_2d(struct)
```

mp-12712

Aims to resemble output of `ase.visualize.plot.plot_atoms` with minor improvements, less conversion hassle and handling of disordered structures.

`pymatgen` is unable to convert disordered structures via `AseAtomsAdaptor().get_atoms(struct)`:

```py
ValueError: ASE Atoms only supports ordered structures
```

which appears to be untrue. ASE does allow for disorder.

**ASE Example**

```py
from pymatgen.ext.matproj import MPRester
from pymatgen.io.ase import AseAtomsAdaptor
from ase.visualize.plot import plot_atoms

struct = MPRester().get_structure_by_material_id('mp-12712')
atoms = AseAtomsAdaptor().get_atoms(struct)

plot_atoms(atoms)
```

ase-mp-12712

**Disordered Example**

```py
from pymatgen.transformations.standard_transformations import SubstitutionTransformation


struct = MPRester().get_structure_by_material_id("mp-19017")

disord_struct: Structure = SubstitutionTransformation(
    {"Fe": {"Fe": 0.75, "C": 0.25}}
).apply_transformation(struct)

plot_structure_2d(disord_struct, rotation="10x,10y,0z", atomic_radii=0.5)
```

mp-19017-disord


janosh:
Structure with new `annotate_sites=True` option:
struct-2d-mp-19017-disordered

Room for improvement: element labels aren't yet occluded by sites with higher z-index.

janosh:
Output seems fine across most of 20 random structures. I think the code can still be simplified quite a bit though so leaving WIP status.

mp-structures-2d

CompRhys:
Maybe a key off to the side? So then it doesn't overlayer the labels?

Does it give different colours to charge assigned elements?

janosh:
> Maybe a key off to the side? So then it doesn't overlayer the labels?

Could you be more specific?

> Does it give different colours to charge assigned elements?

It doesn't. Do you think that's an important feature? I feel like there aren't enough colors to differentiate elements clearly as it. Maybe better to include oxidation state in site labels?

sgbaird:
@janosh if you decide to include it, I think as a site label might work. As a legend (e.g. blue: Ni, red: O) incorporating charge states is tough because of the color issue.

PR #45, opened 2022-08-10
fix: add .py to readme link

sgbaird:


janosh:
Thanks, good catch! 👍 

I should probably add a link checker in CI.

PR #82, opened 2023-06-09
Tiny doc update in ptable.py

JaGeo:
@janosh , just a tiny change to the documentation. I just came accross it.


janosh:
Thanks @JaGeo, good catch! 👍

PR #120, opened 2024-01-09
Minor format tweaks

### LINKED ISSUE #118: [Feature] [Self-assigned] Add heatmap plotter
Hi @janosh, I happen to have some plotting scripts for my MPhil research which includes plotting heatmaps. Shall I proceed and try to merge them into your package?

For example:
headmap1

And this (I didn't find a good way to fix these overlapping labels though):
heatmap2

DanielYang59:
~~Early PR for adding heatmap to resolve #118.~~


pymatviz/utils.py
```diff
@@ -299,23 +299,24 @@ def annotate_metrics(
 
 def crystal_sys_from_spg_num(spg: int) -> CrystalSystem:
     """Get the crystal system for an international space group number."""
-    # not using isinstance(n, int) to allow 0-decimal floats
```
DanielYang59, line 302: Should the case of `spg` not being `int` even be considered here? As the type hint already explicitly requires `type(spg)` to be `int`.

pymatviz/utils.py
```diff
@@ -299,23 +299,24 @@ def annotate_metrics(
...
-    if spg < 75:
+    elif spg <= 74:
         return "orthorhombic"
-    if spg < 143:
+    elif spg <= 142:
         return "tetragonal"
-    if spg < 168:
+    elif spg <= 167:
         return "trigonal"
-    if spg < 195:
```
DanielYang59, line 316: Some nitpicking changes while I'm browsing across the src to get myself familiar with the project. I personally think using a closing range might be cleaner (to read) though should function the same. What do you think?

pymatviz/io.py
```diff
@@ -109,13 +110,13 @@ def save_fig(
         if path.lower().endswith(".svelte"):
             # insert {...$$props} into top-level div to be able to post-process and
             # style plotly figures from within Svelte files
-            with open(path) as file:
+            with open(path, encoding="utf-8") as file:
```
janosh, line 113: I think `encoding="utf-8"` can be [considered the default](https://docs.python.org/3/library/locale.html#locale.getencoding). I would only specify if we encounter a case where it matters.

pymatviz/utils.py
janosh: Type hints can be ignored and it's nice to give specific error messages even then. Often people feed in `pandas` dataframes columns and aren't aware what datatype they have.

pymatviz/__init__.py
```diff
@@ -121,15 +121,14 @@
...
-defaults settings for plotly and matplotlib like increasing font size, prettier
-axis labels (plotly only) and higher figure resolution (matplotlib only).
+# Importing this module has side-effects that apply sensible (often, not always) global
+# defaults settings for plotly and matplotlib like increasing font size, prettier
+# axis labels (plotly only) and higher figure resolution (matplotlib only).
 
-To use it, simply import this module before generating any plots:
+# To use it, simply import this module before generating any plots:
+
+# import pymatviz
```
janosh, line 130: This should have been part of the module doc str at the top all along.

pymatviz/io.py
```diff
@@ -11,6 +11,7 @@
 import matplotlib.pyplot as plt
 import plotly.graph_objects as go
 from tqdm import tqdm
+from weasyprint import HTML
```
janosh, line 14: `weasyprint` is not installed by default and hence cannot be imported at module level.

DanielYang59: Didn't expect you to jump in so early...But here you are as always. Thanks for your suggestion. Yes by doing so the we can automatically generate a `requirements.txt` with `pipreqs` to include it by default? I think this would be cleaner than have some conditional import in the middle and prompt the user to install it manually. What do you think?

janosh: It's a single-use dep. I would say not worth having every user install it if most will never need it.

DanielYang59: Make sense to me! I might reverse this part.

pymatviz/io.py
DanielYang59: Sorry I don't mean to include you into the conversation so early (thus it's a draft). But it's great discussing with you as always. I'm still busying clean up my Paper for ArXiv, and am just work on this from time to time...

Not exactly I would say, from the [page](https://docs.python.org/3/library/locale.html#locale.getencoding) you mentioned there is:
> On Android and VxWorks, return "utf-8".
> On Unix, return the encoding of the current [LC_CTYPE](https://docs.python.org/3/library/locale.html#locale.LC_CTYPE) locale. Return "utf-8" if nl_langinfo(CODESET) returns an empty string: for example, if the current LC_CTYPE locale is not supported.
> On Windows, return the ANSI code page.

Given the popularity of Windows OS, it might be better to explicitly declare the encoding for robustness (I have seen quite some weird issues related to encoding as I'm working on Windows and Linux at the same time), what do you think?

Also [`pylint`](https://pylint.readthedocs.io/en/latest/user_guide/messages/warning/unspecified-encoding.html) encourages explicit encoding by giving this warning if not:

> unspecified-encoding / W1514
> Message emitted:
> Using open without explicitly specifying an encoding
> 
> Description:
> It is better to specify an encoding when opening documents. Using the system default implicitly can create problems on other operating systems. See https://peps.python.org/pep-0597/

pymatviz/utils.py
DanielYang59: I usually do one of two things:

1. Assume the ideal case (right data type is passed), usually for priviate func/method not accessed by the user directly. Which might make the code looks clean but risk the robustness.

2. Assume the worst scenario where any irrational input could be passed in, and implement a strict checking (type, value and so on).

In the original code:
https://github.com/janosh/pymatviz/blob/882a6caaa06acbcadbcfc3d8e2d7fe6aa8ce630e/pymatviz/utils.py#L302-L304

This seems to be standing somewhere in the middle. Maybe I should go ahead and switch to strick checking?

janosh: > Maybe I should go ahead and switch to strict checking?

Certainly, stricter checks and better error messages are always welcome.

DanielYang59: Thanks for the input as always @janosh

pymatviz/heatmap-testdata-enthalpy.csv
DanielYang59, line 1: this would be moved to `tests` dir soon

pymatviz/uncertainty.py
```diff
@@ -185,9 +185,7 @@ def get_std_decay(y_true: ArrayLike, y_pred: ArrayLike, y_std: ArrayLike) -> Arr
     # increasing count of the number of samples in each element of cumsum()
     n_inc = range(1, len(abs_err) + 1)
 
-    decay_by_std = abs_err[y_std_sort].cumsum() / n_inc
-
-    return decay_by_std  # noqa: RET504
```
DanielYang59, line 190: I just notice this ruff tag here. Maybe I'm missing something?

pymatviz/heatmap.py
```diff
@@ -0,0 +1,109 @@
...
+import matplotlib
+import matplotlib.pyplot as plt
+import pandas as pd
+from matplotlib.colors import Colormap
+
+from pymatviz.colormaps import blend_two
+
+
+def heatmap(
+    df: pd.DataFrame,
```
DanielYang59, line 17: I didn't use a more general data type here (ArrayLike for example) because I think we might need to separate data processing from plotting, to simplify plotters (instead of accepting a whole lot of possible types from the **plotter** side). If you agree with this, I might proceed and implement this change, what do you think?

Meanwhile it think it might be better to group plotters into a separate module (dir)?

pymatviz/heatmap.py
```diff
@@ -0,0 +1,109 @@
...
+    # Set x/y axes tick labels
+    ax.set_xticks(range(len(df.columns)))
+    ax.set_xticklabels(df.columns, rotation="vertical")
+
+    ax.set_yticks(range(len(df.index)))
+    ax.set_yticklabels(df.index)
+
+    # Set x/y axes labels
+    ax.set_xlabel(x_label)
+    ax.set_ylabel(y_label)
```
DanielYang59, line 78: It's really hard to write a **general and good looking** plotter without hard-coding cosmetic parameters 😓 (font size etc.).

pymatviz/uncertainty.py
janosh: i thought the variable name was slightly informative so kept it but not important

pymatviz/heatmap.py
janosh: I think `pd.DataFrame` is plenty general enough. And as you said, `pymatviz` is not about getting the data into the right format for you, just getting the right data format on a canvas.

pymatviz/heatmap.py
janosh: as long as everything remains easily customizable, i don't see issues with setting sensible defaults for font size, etc.

DanielYang59:
@janosh Ready for review. Anything improvement needed? Thanks!

pymatviz/heatmap.py
```diff
@@ -0,0 +1,91 @@
...
+
+    # Add colorbar
+    if add_cbar:
+        cbar = ax.figure.colorbar(im, ax=ax)
+        cbar.set_label(cbar_label)
+
+    return ax
```
janosh, line 91: `heatmap()` currently looks a bit like a simple wrapper around `plt.imshow()`. i think a good overarching design principle for functions is how much effort does it save to use the function. to do more in this function, we could make it a bit more task-specific. in the context of materials science, this might be naming the function `element_heatmap()` and defaulting the x/y axis tick labels to element symbols (H, He, Li, Be, ...) so that people can pass np.array/pd.DataFrame without needing to have element symbols as column/row indices

pymatviz/colormaps.py
```diff
@@ -0,0 +1,122 @@
...
+import numpy as np
+from matplotlib.colors import Colormap, LinearSegmentedColormap, ListedColormap
+
+
+def combine_two(
+    cmaps: list[Colormap | str],
+    node: float = 0.5,
+    N: int = 256,
+    reverse: bool = False,
+) -> Colormap:
```
janosh, line 16: isn't this very similar to `matplotlib.colors.LinearSegmentedColormap`? do we really need a custom helper function?

pymatviz/colormaps.py
```diff
@@ -0,0 +1,122 @@
...
+        )
+
+
+def truncate(
+    cmap: str | Colormap,
+    start: float,
+    end: float,
+    N: int = 256,
+) -> Colormap:
+    """Truncate a matplotlib Colormap to a specified range.
```
janosh, line 79: same here, i think we could use `colors.LinearSegmentedColormap.from_list` instead (see https://stackoverflow.com/a/18926541)

janosh (changes requested):
thanks! `assets/heatmap-mixing-enthalpy.svg` looks great. i think the code needs some tweaks though

pymatviz/heatmap.py
DanielYang59: Hi @janosh , thanks for reviewing and these suggestions. Yes it's only after I begin to work on this heatmap function that I realized it's simply a wrappler of `imshow()`. 

For me personally the tricky part for this heatmap plot is actually the manipulation (combine/truncate) of colormaps in an intuitive manner (I would talk more about this in the following specific parts). I might go ahead and separate the column/row labels from the single `pd.DataFrame` to make altering those easier.

However I personally want to make this plotter more general and people could pass in any 2D data array? So I didn't introduce too much task specific parameters...

Another thing might be to add an "interpolation ()" to make generation of the below "smooth" heatmap easier, and to allow labelling data points (but I guess that could be also made into a "decorator" like function to wrap the original function to add some points).

A smooth heatmap

pymatviz/colormaps.py
DanielYang59: Yes it is quite similar, only one main difference here: to assign "[nodes](https://matplotlib.org/stable/users/explain/colors/colormap-manipulation.html)" so that the boundary between children colormaps could be adjusted (for example a composite "two in one" colormap might not necessarily have each child taking up 50% of the colormap, but could be adjusted to say 20%/80%). Let me know if there is more straightforward way to achieve this.

pymatviz/colormaps.py
```diff
@@ -11,18 +11,18 @@
 def combine_two(
     cmaps: list[Colormap | str],
     node: float = 0.5,
-    N: int = 256,
```
DanielYang59: In fact this terrible naming is taken from the official `matplotlib` [doc](https://matplotlib.org/stable/api/_as_gen/matplotlib.colors.Colormap.html) here, just to be consistent with the upstream `matplotlib`, I personally would be more than happy to switch to a more meaningful name as this one.

> ```python
> def __init__(self, name, N=256):
>     """
>     Parameters
>     ----------
>     name : str
>         The name of the colormap.
>     N : int
>         The number of RGB quantization levels.
>     """
> ```

pymatviz/utils.py
```diff
@@ -300,25 +300,26 @@ def annotate_metrics(
...
+    if spg <= 74:
         return "orthorhombic"
-    elif spg <= 142:
+    if spg <= 142:
         return "tetragonal"
-    elif spg <= 167:
+    if spg <= 167:
         return "trigonal"
-    elif spg <= 194:
+    if spg <= 194:
```
DanielYang59, line 320: Interesting change here... I know this should work exactly the same, but I'm just curious why preferring a chain of `if` over `elif`?

For me I prefer `elif` for two reasons:
1. It's a straight forward indicator that exactly one branch would be picked.
2. In other cases where the loop is not broken directly by `return`, use a chain of `if` would lead to the following statements still being assessed even when the former condition has already be met, discussed [here](https://stackoverflow.com/questions/23230044/if-and-elif-chain-versus-a-plain-if-chain), for example:
```python
a = 1

if a > 0:  # this would be assessed
    print("a greater than zero!")
if a < = 0:  # this would also be assessed
    print("a not greater than zero!")
```

pymatviz/colormaps.py
janosh: take your time to recover! no rush

DanielYang59: Hi @janosh I'm back and feel good. This `cmasher` package looks good and is able to handle custom colormaps, so I would drop the `truncate` function, and also see if I could find an already available package for the `combine_two` function as well.

```python
import cmasher as cmr

# Create another custom cmap from custom cmap
built_in_cmap = plt.get_cmap('viridis')
custom_cmap = cmr.get_sub_cmap('viridis', 0.1, 0.9)

new_custom_cmap = cmr.get_sub_cmap(custom_cmap, 0.1, 0.3)

# Generate some data to plot
x = np.random.rand(100)
y = np.random.rand(100)
z = x**2+y**2

# Make scatter plot of data with colormap
plt.scatter(x, y, c=z, cmap=new_custom_cmap, s=300)
plt.show()
```

pymatviz/colormaps.py
DanielYang59: I'm trying to find an existing function for this and may remove this part of code after that😊.

DanielYang59:
I notice `matplotlib` already has the `interpolation` method built in to allow interpolation of the plot, as stated [here](https://matplotlib.org/stable/gallery/images_contours_and_fields/interpolation_methods.html), by `ax.imshow(grid, interpolation=interp_method, cmap='viridis')`.

In this case I think this entire code looks pointless, being a simple wrapper over `imshow`. Therefore I wish to remove this entire function altogether. However the heatmap itself looks good to me (it might not make sense to leave a standalone heatmap tutorial because `pymatviz` is a visualization package instead of a tutorial for `matplotlib`?)...Do you have any suggestions @janosh? Thanks.

pymatviz/colormaps.py
DanielYang59: I might go ahead and PR this function to some low level python package (`cmasher` for example), and remove `combine_two` from here after that, to keep `pymatviz` central to high level ops.

DanielYang59:
@janosh Can you please review this as a minor format tweak and I would open another PR for the periodic table scatter plotter? Thanks!

pymatviz/utils.py
janosh: sorry, forgot to respond.

i'd be happy to use `elif` too but `ruff` complains and didn't disagree enough to want to override `ruff`'s config

janosh (approved):
thanks @DanielYang59! better comments and more tests are always appreciated! 👍

DanielYang59:
Thanks @janosh! I would also add guides and assert generations for `ptable_hist` as well in the following PR.

pymatviz/utils.py
DanielYang59: Maybe I would see if I could tag `ruff` to ignore this part in the following PR! Thanks!

DanielYang59:
Meanwhile I pushed the `combine_colormap` function to upstream `CMasher` at 1313e/CMasher/pull/122. Hopefully it could get accepted 😁.

DanielYang59:
> Meanwhile I pushed the `combine_colormap` function to upstream `CMasher` at [1313e/CMasher/pull/122](https://github.com/1313e/CMasher/pull/122). Hopefully it could get accepted 😁.

A quick update on this @janosh , after a month's review it finally get merged (really learned a lot about details in coding during this process), and it's now available as, detailed in the [manual](https://cmasher.readthedocs.io/user/usage.html#combine-colormaps):
```python 
from cmasher import combine_cmaps

cmr.combine_cmaps(cmap_0, cmap_1)
```

janosh:
nice, glad you found a good home for it! great team work between you and the maintainers on that PR 👍

Issue #130, opened 2024-04-05
`ptable_heatmap` with diagonally-split tiles


keeganq:
Hi, thanks for this really awesome repository! Very easy to use with satisfying results.

Are there plans to add functionality for joint periodic table plots for comparing two values?  Similar to the one shown here: https://www.nature.com/articles/s42256-023-00716-3/figures/2

Would be great to have this kind of visualization in the toolkit, as I think it's much easier than trying to compare two `ptable_heatmap` plots.

janosh:
yeah, i almost implemented this as a reusable function at one point. don't remember why it didn't happen and unfortunately don't have time right now. but it's been on the roadmap for a while and a PR would be very welcome!

pinging @DanielYang59 just in case you'd like to work on this. no expectations of course

DanielYang59:
Looks good. I would have a try and see how everything goes. Thanks for pinging me!

DanielYang59:
Hi @keeganq We just added the requested plotter with #131, and please let me know if you have further suggestions on existing plotter.

To use the new plotter, you could use the following (basically just pass a `{element_name: [*values]}` dict to the plotter):
https://github.com/janosh/pymatviz/blob/492cd5301f9b241291a49f24a4d8adbbd50ae5c6/examples/_generate_assets.py#L218-L232

keeganq:
This is awesome, @DanielYang59. Thank you!

DanielYang59:
No worries at all. This can never be done so well without @janosh's suggestions. 

Meanwhile we're trying to improve the API for ptable plotters, so if you have any comments/suggestions during usage, please let me know. That would be valuable feedback for me.

PR #131, opened 2024-04-06
Refactor `ptable` plotters and add `ptable_heatmap` with diagonally-split tiles

DanielYang59:
closes https://github.com/janosh/pymatviz/issues/130

### Good news first
- New implementation for plotting nested ptable of course :) everything is now modular, I'm feeling very happy.
- The requested evenly-split ptable :)

### Breaking changes
- `ptable_plots` replaced by `ptable_lines` and `ptable_scatters` with the new implementation, kind of good news :) 
- Use `*_kwargs` over `*_kwds` for consistency, previously a random mixture of both were used.

### Urgently needed fixes:
- Current implementation of `ptable_lines` and `ptable_scatters` doesn't support colormap (yet). I would certainly implement the support for `ptable_scatters`, but as we discussed previously in https://github.com/janosh/pymatviz/pull/131#issuecomment-2058143028, I'm not sure if we want to support colored symbols for `ptable_lines`.
- `mace_pair_repulsion.py` is certainly broken as colormap is not supported by `ptable_lines` anymore.

### Not-so-urgent fixes/enhancement:
- `data_preprocessor` doesn't support missing value/infinity handling yet.
- Replace other ptable plotters with new implementation.
- Finalize data preprocessing tool
- Add instruction for custom child plotters


pymatviz/structure_viz.py
```diff
@@ -410,8 +410,10 @@ class used to draw chemical bonds. Allowed are edgecolor, facecolor, color,
         ):
             try:
                 struct.add_oxidation_state_by_guess()
-            except ValueError:  # fails for disordered structures
-                "Charge balance analysis requires integer values in Composition"
+            except ValueError as exc:  # fails for disordered structures
+                raise ValueError(
+                    "Charge balance analysis requires integer values in Composition"
```
DanielYang59, line 415: This change leads to multiple failures in unit tests, I'm not sure what is intended for this part? Suppress the `ValueError` or re-raise an `ValueError` with custom message?

pymatviz/powerups.py
```diff
@@ -196,12 +196,10 @@ def annotate_metrics(
         "MSE": lambda x, y: ((x - y) ** 2).mean(),
         "MAPE": mape,
         "R2": r2_score,
-        # TODO: check this for correctness
```
DanielYang59, line 199: I think the formula for adjusted R2 is correct, as per [the reference](https://www.statisticshowto.com/probability-and-statistics/statistics-definitions/adjusted-r2/) here:
image

In our case the `k` (the number of independent regressors, i.e. the number of [variables](https://www.statisticshowto.com/probability-and-statistics/types-of-variables/) in your model) is 1 (single y only)

pymatviz/utils.py
```diff
@@ -72,10 +72,10 @@ def pretty_label(key: str, backend: Backend) -> str:
     return symbol_mapping.get(key, {}).get(backend, key)
 
 
-def crystal_sys_from_spg_num(spg: int) -> CrystalSystem:
+def crystal_sys_from_spg_num(spg: float) -> CrystalSystem:
```
DanielYang59, line 75: Use `float` over `int` as we intend to accept either `integer or float with no decimal part`

DanielYang59:
Can you please give me some comments so far @janosh ? Thanks a lot!

assets/ptable-diags.svg
janosh, line 1: wow, this looks excellent! 👍

pymatviz/ptable.py
```diff
@@ -160,6 +162,231 @@ def count_elements(
     return srs
 
 
+def ptable_diags(
```
janosh, line 165: maybe long-term we want to support more than 2 triangles per tile (i.e. 4)? in that case, maybe it makes sense to give this a more general name like `ptable_heatmap_split_tiles`? alternatively, maybe this shouldn't be a separate function but integrated into `ptable_heatmap()`? then the function would switch behavior based on the number of values per-element and raise ValueError if it's more than 4.

pymatviz/ptable.py
```diff
@@ -160,6 +162,231 @@ def count_elements(
...
+                    anno_text = anno_text(plot_data)
+                annotation["text"] = anno_text
+            ax.annotate(**(defaults | annotation))
+
+        if plot_data is not None:
+            # Fill areas above and below the line
+            x = np.linspace(0, 10, 10)
+            y = np.linspace(10, 0, 10)
+            ax.fill_between(x, y, 10, color=cmap(scaled_data[symbol][0]))
+            ax.fill_between(x, y, 0, color=cmap(scaled_data[symbol][1]))
```
janosh, line 349: this is smart! 👍 i like this way of making the triangles. on the other hand, it could be finicky to support more than 2 triangles with this like [in this example](https://stackoverflow.com/a/66050636)?
O3Pbi

janosh (commented):
impressive work! thanks so much @DanielYang59. 🙏 several comments below

i think the SVG compression action is failing because of this unaddressed limitation
https://github.com/actions/checkout/issues/694

pymatviz/ptable.py
DanielYang59: Yes it makes sense to merge it into heatmap but I would need to think about how to switch in this case.

I thought about supporting arbitrary/dynamic number of splits (depending on the length of the list passed in for each element) actually, even allowing rotating the separation line (image a circle being split into several parts and then cropped as a rectangle)...But you know, could think of doesn't mean could implement, so I ended up limiting to only two. 

But it certainly would be good to add more options (4 for example), I would implement this tomorrow.

Meanwhile ptable plotters have heavily repeated code and I'm not sure if we could do something about it.

assets/ptable-diags.svg
DanielYang59: Thanks!

pymatviz/ptable.py
DanielYang59: Yes the 4-split method  looks very geeky. But it's certainly good to have the function. I believe we could do better with our line-split approach. 

I would have a try and implement a helper function to split a rectangle into even parts tomorrow.

pymatviz/ptable.py
janosh: > Meanwhile ptable plotters have heavily repeated code and I'm not sure if we could do something about it.

that's true. i've been putting off cleaning this up for a while. i think the simplest way forward would be to identify common chunks of code and extract into a helper function that is called by the different `ptable_heatmaps`.

pymatviz/ptable.py
DanielYang59: I'm playing around with a more clean yet powerful (helpfully) method: build a pie chart with the `matplotlib` standard `pie` API, and crop a rectangle out of it. In this way we could have any number of splits and don't need to hard code the separation lines, but I'm still playing around with the following toy code before integrating it into the `ptable` plotter:

```python
"""Toy code to split a rectangle into equal parts."""


import numpy as np
from matplotlib.colors import Colormap, Normalize


def plot_split_rectangle(
    colors: list,
    start_phase: float = 0,  # TODO: control rotation
    ) -> None:
    # Plot the pie chart
    fig, ax = plt.subplots(figsize=(6, 6))
    ax.pie(np.ones(len(colors)), colors=colors)

    # Crop a central rectangle from the pie chart
    path = plt.Rectangle((-0.5, -0.5), 1, 1, fc='none', ec='none')
    ax.add_patch(path)

    ax.set_xlim(-0.5, 0.5)
    ax.set_ylim(-0.5, 0.5)

    # Hide axes spines
    ax.spines['top'].set_visible(False)
    ax.spines['right'].set_visible(False)
    ax.spines['bottom'].set_visible(False)
    ax.spines['left'].set_visible(False)

    # Hide axes ticks
    ax.tick_params(axis='both', which='both', bottom=False, top=False, left=False, right=False)

    ax.set_aspect('equal', adjustable='box')  # not seem to work?

    plt.show()

if __name__ == "__main__":
    # Prepare colors from Colormap
    data = [0, 10, 20, 30, 40, 60]

    cmap: Colormap = plt.get_cmap("coolwarm")

    # Normalize data
    norm = Normalize(vmin=min(data), vmax=max(data))
    normalized_data = [norm(datum) for datum in data]

    colors = [cmap(data) for data in normalized_data]

    # Test plotter
    plot_split_rectangle(colors)

```

Which gives: 
image

assets/ptable-diags.svg
DanielYang59, line 1: @janosh Look at that! We're getting really close! We could even do different split in a single plot (some element could split into N, others might do M, completely dependent on the data length).

But for some reason the center cropping is not working. Also we need to merge this into the heatmap plotter.

janosh: very nice! 👍  we could give [`Axes.set_clip_path`](https://matplotlib.org/stable/gallery/images_contours_and_fields/image_clip_path.html) a try. [it's supposed to work for `Patches` too](https://matplotlib.org/3.1.1/api/_as_gen/matplotlib.axes.Axes.set_clip_path.html), not just images as

DanielYang59: That is exactly what I was reading. Thanks!

DanielYang59: For some reason cropping doesn't work. Do you have time to give me a hand on this? Thanks!

janosh: wow, this is really bad API design! you have to [pass a non-obvious override](https://matplotlib.org/stable/api/_as_gen/matplotlib.axes.Axes.pie.html) to make clipping work:

```diff
ax.pie(
    np.ones(len(colors)),
    colors=colors,
    startangle=start_angle,
+  wedgeprops=dict(clip_on=True),
)
```

then it looks really nice! 🎉 

ptable-diags

maybe default the `start_angle: float = 45`?

pymatviz/ptable.py
DanielYang59: > i think the simplest way forward would be to identify common chunks of code and extract into a helper function that is called by the different `ptable_heatmaps`

I know it's easier said that done, maybe we should try to separate them into the periodic table itself and the nested plotter. I might have a go at this later.

assets/ptable-diags.svg
DanielYang59: Bravo!!!!!

Yes of course the default `start_angle` should be changed. But if we want to recreate the original issue, we then need to set it as 135.

DanielYang59:
I think we're almost there @janosh. A few issues left:
- I just tried to merge this plotter into heatmap but only to find the final code very messy and confusing. The `heatmap` plotters have extensive code to process single values, if we then make it handle `float | list[float]`, the code seems too complex to maintain. I would suggest still having them side by side (the new plotter could handle no-split case just fine) .
- I used a random length data in `asset` to showcase that this new plotter could handle arbitrary separation, but the final plot seems quite messy. Maybe still use 2-split?

pymatviz/ptable.py
```diff
@@ -1174,11 +1175,11 @@ def ptable_plots(
                 scatter_kwargs = plot_kwds or {}
 
             # Plot without colormap when data len is 2
-            if plot_data.ndim == 2:
+            if len(plot_data) == 2:
                 ax.plot(plot_data[0], plot_data[1], **(scatter_kwargs or {}))
 
             # Plot with colormap when data len is 3
-            elif plot_data.ndim == 3:
+            elif len(plot_data) == 3:
```
DanielYang59, line 1182: I'm a little confused what we were trying to achieve here:
```python
            # Plot with colormap when data len is 3
            elif len(plot_data) == 3:
                ax.plot(
                    plot_data[0],
                    plot_data[1],
                    c=plot_data[2],
                    cmap=cmap,
                    **(scatter_kwargs or {}),
                )
```
The `plot_data[2]` should be a array the same length as `plot_data[0/1]`. But I think we could only assign a single color for a line (unlike the case for scatters?)?

janosh:
> I would suggest still having them side by side (the new plotter could handle no-split case just fine) .

given the new plotter can handle both cases, can we not make it replace the old implementation? that would allow keeping all the existing tests for `ptable_heatmap` to make sure we don't regress in any way. it would just be an alias then

```py
ptable_heatmap = ptable_splits
```

> I used a random length data in asset to showcase that this new plotter could handle arbitrary separation, but the final plot seems quite messy. Maybe still use 2-split?

let's make 2 assets that show the 2 and 3 splits case. people will guess that 4 also works and having different split numbers per element seems very rare.

> How to compress the svg? It added thousands of lines. (thought you did it before)

```
svgo --multipass path/to/file.svg
```

is very convenient. the best level of compression i've seen is offered by https://vecta.io/nano which unfortunately don't offer a free API so you have to manually upload files there.

> We might need to handle special values if we allow single value (like NaN or infinity)? This sounds like something heatmap should handle, do you want to prevent user from giving such values to the new plotter?

yes, let's support bad values. we should try to do sensible data cleaning as much as possible. like pandas, ideally we rarely error but just ignore NaNs and maybe plot infinities with a special color (e.g. gray or white or black). but that could wait for a future PR

pymatviz/ptable.py
janosh: yes, this is the issue i noticed in https://github.com/janosh/pymatviz/issues/132#issuecomment-2049115252. specifically what i meant when i wrote

> this should be `if len(plot_data) == 2:` and `if len(plot_data) == 3:` instead of checking `ndim`. 🤦
> 
> the annoying thing is, `ax.plot()` only supports a `color` (or `c`) as a single string, not different colors as a list of values for each data point. so we need to go back to `ax.scatter` for the `len=3` case. but then the `scatter_kwargs` are broken since `ax.plot` and `ax.scatter` have very different keywords. quite annoying.

i don't have a good solution yet. i briefly tried changing the color of each data point afterwards like this

```py
for point, color in zip(zip(*plot_data[:2]), plot_data[2]):
    ax.plot(*point, color=cmap(color), **scatter_kwargs)

```

but don't think this will work and would be a bad solution anyway

pymatviz/ptable.py
DanielYang59: I just opened #133 for more detailed discussion on this :)

pymatviz/ptable.py
DanielYang59: I would fold this for now because I'm confident this problem would be gone if we could manage to achieve #133 😃

DanielYang59:
> given the new plotter can handle both cases, can we not make it replace the old implementation? \

Yes current implementation is more versatile in plotting. But the old implementation has quite some data processing part. I'm hoping we could gradually clarify everything with #133 and #134. But it's certainly a lot of work ahead.

> > I used a random length data in asset to showcase that this new plotter could handle arbitrary separation, but the final plot seems quite messy. Maybe still use 2-split?
> 
> let's make 2 assets that show the 2 and 3 splits case. people will guess that 4 also works and having different split numbers per element seems very rare.

That's a good idea (thought we should avoid duplicate of the same plotter)!

> ```
> svgo --multipass path/to/file.svg
> ```
> 
> is very convenient. the best level of compression i've seen is offered by https://vecta.io/nano which unfortunately don't offer a free API so you have to manually upload files there.

Thanks a lot for the input.

> > We might need to handle special values if we allow single value (like NaN or infinity)? This sounds like something heatmap should handle, do you want to prevent user from giving such values to the new plotter?
> 
> yes, let's support bad values. we should try to do sensible data cleaning as much as possible. like pandas, ideally we rarely error but just ignore NaNs and maybe plot infinities with a special color (e.g. gray or white or black). but that could wait for a future PR

I believe this could be resolved by #134 too 😄

DanielYang59:
I'm not sure if `add_element_type_legend` is necessary for `ptable` heatmaps. Because heatmaps are intended to color ptable by its value, and therefore alters the default color of each element type block.

janosh:
i think we don't need `add_element_type_legend`. changing the background color of tiles definitely doesn't make sense. the color of the element symbol maybe but i think even that becomes visually too noisy

DanielYang59:
> i think we don't need `add_element_type_legend`. changing the background color of tiles definitely doesn't make sense. the color of the element symbol maybe but i think even that becomes visually too noisy

Yes I'm thinking the same. I would remove them then. I'm beginning to sketching the "ptable" projector and I'm not having a solid issue in mind now. So free feel to push your great ideas!

DanielYang59:
Current API looks like something user has to call multiple separate methods in order to get the final ptable plot, something like:
```python

# Initialize plotter with some global style settings
plotter = PTableProjector(colormap, **kwargs)

# Somehow initialize a child plotter to be projected/nested which I don't know how to achieve now
child_plotter = no_idea_now
plotter.add_child_plots(child_plotter)

plotter.add_colorbar(args)

plotter.add_ele_symbols(args)
```

And then we could either:
1. Use the original API to wrap up the new projector to avoid breaking changes. But of course I'm hoping the new methods would reduce maintenance burden.
2. Expose the new API directly to users and let them do things step by step, just like above.

I would personally prefer the latter because I think packing all args into a single function call is a bit confusing (even read the huge docstring is hard for me), and separate them would make it clearer which args correspond to what functionality.

Anyway it's something easy to change and we could focus on the implementation now until we're ready to merge 😃

assets/ptable_new.svg
DanielYang59, line 1: @janosh. Good news! Finally get the completely refactored ptable plotter to work!

janosh: glad this is coming along :)

janosh:
> Current API looks like something user has to call multiple separate methods in order to get the final ptable plot, something like:
> 
> ```python
> # Initialize plotter with some global style settings
> plotter = PTableProjector(colormap, **kwargs)
> 
> # Somehow initialize a child plotter to be projected/nested and pass it to the projector
> # But I don't know how to achieve this now
> child_plotter = no_idea_now
> plotter.add_child_plots(child_plotter)
> 
> plotter.add_colorbar(**cbar_args)
> 
> plotter.add_ele_symbols(**symbol_args)
> ```

really appreciate you exploring new ideas for the `pymatviz` API! 👍 
i'm not super keen on making it object-oriented though. i quite like for plots to be pure functions. i think some of the quirks and complications from working with `matplotlib` come from their overly OOP API which has tons of state and often has too many different places where settings/config must be passed and that need to be checked when something goes wrong. i wonder if there's anything we actually need an OOP interface for in pymatviz that can't be done with pure functions?

DanielYang59:
> really appreciate you exploring new ideas for the `pymatviz` API! 👍 

All good. Enjoy this process very much.

> i'm not super keen on making it object-oriented though. i quite like for plots to be pure functions. i think some of the quirks and complications from working with `matplotlib` come from their overly OOP API which has tons of state and often has too many different places where settings/config must be passed and that need to be checked when something goes wrong. 

I don't have a strong preference for a specific coding paradigm for now. But now that we're building a higher level API on top of `matplotlib`, it might be hard to completely jump out of its framework. Yes I think the API of `matplotlib` is pretty hard for me, but I guess it's necessary as there are just too many "states"/parameters that could be tweaked to make a better plot. And the results are quite objective, there is no magic formula or such so that we could calculate a set of optimal plotting parameters... which I wish there is...

> i wonder if there's anything we actually need an OOP interface for in pymatviz that can't be done with pure functions?

I'm not actually pushing towards making it OOP though (maybe I don't have a strong understand of OOP vs functional programming). My intention was to make everything modular (grouped by objects of course, colorbar/colormap/element symbol/or such) so that it becomes clearer for users (contrary to packing all args into a single plotter call), and make it easier for us to maintain.

I think by separating everything by object, the args are more self-explanatory. Instead of:
```python
def ptable_splits(
    data: pd.DataFrame | pd.Series | dict[str, list[list[float]]],
    colormap: str,
    start_angle: float = 135,
    ax_kwds: dict[str, Any] | None = None,
    symbol_pos: tuple[float, float] = (0.5, 0.5),
    cbar_coords: tuple[float, float, float, float] = (0.18, 0.8, 0.42, 0.02),
    cbar_title: str = "Values",
    cbar_title_kwds: dict[str, Any] | None = None,
    cbar_kwds: dict[str, Any] | None = None,
    anno_kwds: dict[str, Any] | None = None,
    **kwargs: Any,
) -> plt.Figure:
    """Plot evenly-split tiles, nested inside a periodic table.

    Args:
        data (pd.DataFrame | pd.Series | dict[str, list[list[float]]]):
            Map from element symbols to plot data. E.g. if dict,
            {"Fe": [1, 2], "Co": [3, 4]}, where the 1st value would
            be plotted on the lower-left corner and the 2nd on the upper-right.
            If pd.Series, index is element symbols and values lists.
            If pd.DataFrame, column names are element symbols,
            plots are created from each column.
        colormap (str): Matplotlib colormap name to use.
        start_angle (float): The starting angle for the splits in degrees,
                and the split proceeds counter-clockwise (0 refers to
                the x-axis). Defaults to 135 degrees.
        ax_kwds (dict): Keyword arguments passed to ax.set() for each plot.
            Use to set x/y labels, limits, etc. Defaults to None. Example:
            dict(title="Periodic Table", xlabel="x-axis", ylabel="y-axis", xlim=(0, 10),
            ylim=(0, 10), xscale="linear", yscale="log"). See ax.set() docs for options:
            https://matplotlib.org/stable/api/_as_gen/matplotlib.axes.Axes.set.html#matplotlib-axes-axes-set
        symbol_text (str | Callable[[Element], str]): Text to display for
            each element symbol. Defaults to lambda elem: elem.symbol.
        symbol_kwargs (dict): Keyword arguments passed to plt.text() for
            element symbols. Defaults to None.
        symbol_pos (tuple[float, float]): Position of element symbols
            relative to the lower left corner of each tile.
            Defaults to (0.5, 0.5). (1, 1) is the upper right corner.
        cbar_coords (tuple[float, float, float, float]): Colorbar
            position and size: [x, y, width, height] anchored at lower left
            corner of the bar. Defaults to (0.25, 0.77, 0.35, 0.02).
        cbar_title (str): Colorbar title. Defaults to "Values".
        cbar_title_kwds (dict): Keyword arguments passed to
            cbar.ax.set_title(). Defaults to dict(fontsize=12, pad=10).
        cbar_kwds (dict): Keyword arguments passed to fig.colorbar().
        anno_kwds (dict): Keyword arguments passed to plt.annotate()
            for element annotations. Defaults to None. Useful for adding
            e.g. number of data points in each plot. For that, use
            anno_kwds=lambda plot_vals: dict(text=len(plot_vals)).
            Recognized keys are text, xy, xycoords, fontsize, and any other
            plt.annotate() keywords.
        on_empty ('hide' | 'show'): Whether to show or hide tiles for elements without
            data. Defaults to "hide".
        **kwargs: Additional keyword arguments passed to plt.subplots().
```

We now have (sorry it's not a fair comparison, I still haven't finished all docstrings):
```python
class PTableProjector:
    """Project (nest) a custom plot into a periodic table."""

    def __init__(
        self,
        colormap: str | Colormap,
        data: SupportedDataType,
        **kwargs: Any,
    ) -> None:
        """Initialize a ptable projector.

        Default figsize is set to (0.75 * n_groups, 0.75 * n_periods).

        Args:
            colormap (str | Colormap): The colormap to use.
            data (SupportedDataType): The data to be visualized.
            **kwargs (Any): Additional keyword arguments to
                pass to the plt.subplots function call.
        """

def add_child_plots(
        self,
        child_plotter: Callable[[plt.axes, *Any], None],
        child_args: dict[str, Any],
    ) -> None:
        """Add selected custom child plots."""

def add_ele_symbols(
        self,
        text: Callable[[Element], str] = lambda elem: elem.symbol,
        pos: tuple[float, float] = (0.5, 0.5),
        kwargs: dict[str, Any] | None = None,
    ) -> None:

def add_colorbar(
        self,
        title: str,
        coords: tuple[float, float, float, float] = (0.18, 0.8, 0.42, 0.02),
        cbar_kwds: dict[str, Any] | None = None,
        title_kwds: dict[str, Any] | None = None,
    ) -> None:
```

Personally I think the latter is less overwhelming (at least for me...).

But anyway, with the new API, we still have the flexibility to pack it up and use just like before :)

DanielYang59:
Maybe you could give me a hand or some comments on this one (if you have time of course, no pressure) @janosh? Feeling a "little bit" (just a bit) overwhelmed by the amount of work left. 

Current progress roughly being:
1. The ptable projector has been merged, and tested to work on the new diagonal-split plotter and line plotter. However the default style of line plotters seems terrible, and args are not passed to child line plotter.
2. The data preprocessor doesn't seem to work well on data for line plotters, where each element has a data of nested lists/array tuple[np.ndarray, np.ndarray]. The original processor is not prepared for this.
3. The other plotters need to be replaced by new implementation of course, but I prefer to work on one each time.

No pressure of course, if you're current busy just ignore this :)

DanielYang59:
@janosh. Can you please review this? Sorry to say but I'm feeling that I'm currently making this PR too huge for me to efficiently to work on. Can we merge this first and I would chop up the remaining work into (multiple) follow-up PRs.

### Good news first
- New implementation for plotting nested ptable of course :) everything is now modular, I'm feeling very happy.
- The requested evenly-split ptable :)

### Breaking changes
- `ptable_plots` replaced by `ptable_lines` and `ptable_scatters` with the new implementation kind of good news :) 

### Urgently needed fixes:
- Current implementation of `ptable_lines` and `ptable_scatters` doesn't support colormap (yet). I would certainly implement the support for `ptable_scatters`, but as we discussed previously in https://github.com/janosh/pymatviz/pull/131#issuecomment-2058143028, I'm not sure if we want to support colored symbols for `ptable_lines`.
- `mace_pair_repulsion.py` is certainly broken as colormap is not supported by `ptable_lines` anymore.

### Not-so-urgent fixes/enhancement:
- `data_preprocessor` doesn't support missing value/anomaly handling yet.
- replace other ptable plotters with new implementation.

DanielYang59:
@janosh In case this PR get lost. Thanks for your time!

janosh:
sorry for the delay. i'll try to get on this today.

DanielYang59:
> sorry for the delay. i'll try to get on this today.

No worries at all. No pressure. Thanks!

pymatviz/ptable.py
```diff
@@ -785,9 +1187,6 @@ def ptable_hists(
     log: bool = False,
     anno_kwds: dict[str, Any] | None = None,
     on_empty: Literal["show", "hide"] = "hide",
-    color_elem_types: Literal["symbol", "background", "both", False]
-    | dict[str, str] = "background",
-    elem_type_legend: bool | dict[str, Any] = True,
     **kwargs: Any,
```
janosh, line 1190: maybe a misunderstanding. [when i said](https://github.com/janosh/pymatviz/pull/131#issuecomment-2058143028)

> i think we don't need `add_element_type_legend`. changing the background color of tiles definitely doesn't make sense. the color of the element symbol maybe but i think even that becomes visually too noisy

i was referring to the `ptable_splits` function. i think for `ptable_hists` these arguments make more sense, probably enough to keep them.

pymatviz/ptable.py
```diff
@@ -1039,238 +1395,221 @@ def ptable_plots(
...
-        # Get the min/max values for norm calculation
-        third_lists = [data[2] for data in data.values()]
-        vmin = min(min(third_list) for third_list in third_lists)
-        vmax = max(max(third_list) for third_list in third_lists)
+    return plotter.fig
 
-        norm = Normalize(vmin=vmin, vmax=vmax)
 
-        cbar_ax = fig.add_axes(cbar_coords)
+def ptable_splits(
```
janosh, line 1511: maybe rename this to `ptable_heatmap_splits` so users immediately know what is being split?

janosh (commented):
i think everything in this PR looks great. you left some todos which IIUC you prefer to leave for subsqeuent PRs?

> My intention was to make everything modular (grouped by objects of course, colorbar/colormap/element symbol/or such) so that it becomes clearer for users (contrary to packing all args into a single plotter call), and make it easier for us to maintain.
>
> I think by separating everything by object, the args are more self-explanatory. Instead of:

i take your point! i'm still a bit reluctant about the `PTableProjector` class but maybe you're right. i was originally thinking to extract common logic between `ptable_...` functions into helper functions like `data_preprocessor` and `handle_missing_and_anomaly` and then give each `ptable_...` function kwargs for each helper function:

```py
ptable_heatmap(data, data_preproces_kwargs=dict(...), handle_missing_kwargs=dict(...), ...)
```

to simplify the function signature and make it less overwhelming. users could jump between the helper functions to only see the keywords they currently care about while still leaving working only with pure functions. maybe worth considering if `PTableProjector` could be rewritten that way?

pymatviz/ptable.py
DanielYang59: Yes! Such a wonderful idea.

pymatviz/ptable.py
DanielYang59: Sure! I would revert it soon.

DanielYang59:
> i think everything in this PR looks great. you left some todos which IIUC you prefer to leave for subsqeuent PRs?

Yes. There are too many changes to make for a single PR, I would address them shortly in a follow up PR.

> > My intention was to make everything modular (grouped by objects of course, colorbar/colormap/element symbol/or such) so that it becomes clearer for users (contrary to packing all args into a single plotter call), and make it easier for us to maintain.
> > I think by separating everything by object, the args are more self-explanatory. Instead of:
> 
> i take your point! i'm still a bit reluctant about the `PTableProjector` class but maybe you're right. i was originally thinking to extract common logic between `ptable_...` functions into helper functions like `data_preprocessor` and `handle_missing_and_anomaly` and then give each `ptable_...` function kwargs for each helper function:
> 
> ```python
> ptable_heatmap(data, data_preproces_kwargs=dict(...), handle_missing_kwargs=dict(...), ...)
> ```
> 
> to simplify the function signature and make it less overwhelming. users could jump between the helper functions to only see the keywords they currently care about while still leaving working only with pure functions. maybe worth considering if `PTableProjector` could be rewritten that way?

I understand your intention here and I'm not sure which one would be easier for the user though. I think by packing everything into the `PTableProjector` class, what I was trying to achieve is simply to pack everything together (to make the module look cleaner, instead of having a bunch of helper functions loosely laid everywhere)? I think they should work more or less the same because in my class, there isn't much (if any) share states between different methods at all.

pymatviz/ptable.py
DanielYang59: Reverted.

pymatviz/ptable.py
```diff
@@ -160,6 +214,418 @@ def count_elements(
...
+
+        # Initialize periodic table canvas
+        n_periods = df_ptable.row.max()
+        n_groups = df_ptable.column.max()
+
+        # Set figure size
+        plot_kwargs = plot_kwargs or {}
+        plot_kwargs.setdefault("figsize", (0.75 * n_groups, 0.75 * n_periods))
+
+        self.fig, self.axes = plt.subplots(n_periods, n_groups, **plot_kwargs)
```
janosh, line 361: > I think both should work almost the same because in my class, there isn't much (if any) shared states at all.

i would consider self.(...) attribute as state. but rather than belabor the point, let's merge this PR as is nad refactor later if needed.

janosh (approved):
thanks a lot @DanielYang59, this is a huge PR.

DanielYang59:
Thanks for reviewing! I would try to finish the remaining several TODOs soon.

> > I think both should work almost the same because in my class, there isn't much (if any) shared states at all.
> 
> i would consider any `self.<attribute>` as state. but rather than belabor the point, let's merge this PR as is and refactor later if needed.

Yes but nearly zero states are "shared". So it should be pretty easy if we really wish to refactor it.

PR #139, opened 2024-05-08
Remove text background and fix z-order in `structure_viz`

DanielYang59:
### Summary

- Remove text background 
- Fix z-order such that objects at the foreground would properly occlude those at the background
- BREAKING: remove `site_labels_bbox` argument
- Replace deprecated `MPRester` with `mp_api` API in `_generate_assets.py`

### Before:
image

### After:
struct-2d-mp-19017-Li4Mn0 8Fe1 6P4C1 6O16-disordered


DanielYang59:
@janosh Please review this at your convenience. Thanks!

janosh:
nice work! 👍 definite improvement

> Perhaps we should revise the plotter such that the symbol of elements at the back would be blocked by elements in the front?

maybe we can make it an option whether element symbols should be occluded, e.g. `occlude_labels: bool = true` and include it in this PR? [`zorder`](https://matplotlib.org/3.1.1/gallery/misc/zorder_demo.html) might help here

DanielYang59:
Wonderful. Z order is exactly what we need. And we need to define the `zorder` (perhaps by distance to the viewpoint). I would do this today. Thanks for the motivation.

DanielYang59:
All done. Please review @janosh. Thanks!

pymatviz/structure_viz.py
```diff
@@ -363,29 +379,37 @@ class used to draw chemical bonds. Allowed are edgecolor, facecolor, color,
...
                     # disordered sites
                     half_way = 2 * np.pi * (start + occupancy / 2)
                     direction = np.array([math.cos(half_way), math.sin(half_way)])
                     text_offset = (
                         (0.5 * radius) * direction if occupancy < 1 else (0, 0)
                     )
 
-                    bbox = dict(facecolor="none", edgecolor="none", pad=1)
-                    bbox.update(site_labels_bbox or {})
+                    bbox = dict(facecolor="none", edgecolor="none", pad=1, alpha=0)
```
DanielYang59, line 392: TODO soon in this PR: drop `bbox` altogether

assets/matbench-phonons-structures-2d.svg
janosh, line 1: i think this file needs regenerating. the oxidation states are gone

janosh (approved):
thanks @DanielYang59! 👍

assets/matbench-phonons-structures-2d.svg
DanielYang59: Sorry I didn't notice this, yes as I mentioned here https://github.com/janosh/pymatviz/pull/139#issuecomment-2099727351, the fetching behavior of oxidation states still inconsistent for some reason.

PR #140, opened 2024-05-08
Add keyword `hide_f_block: bool = None` (La and Ac series) to `ptable` plotters

### LINKED ISSUE #138: Several possible improvements for `ptable_heatmap` plotter
I noticed several minor issues with `ptable_heatmap` plotter:
- The font style of the text in each tile and colorbar title cannot be controlled individually (only a single `text_style` arg is allowed). (Maybe the new implementation would fix this automatically because they would be separate objects then)
- The colormap is named as `colorscale`, maybe rename it to `colormap` for consistency with `matplotlib` and other parts of the code?
- The default setting doesn't format numbers in (-1, 1) range well, see the below example (data for Ag and Zn is 0.93 and -0.43). Manually set `fmt` fixes the issue though.

  test_plot

- Should we add an option to allow easy hiding of the Lanthanum and Actinium series (maybe there is already one that I'm not aware of)?

DanielYang59:
## Summary

- Allow easy hiding of f-block (La and Ac series) in ptable plotter, to close #138

- (**Need discussion**) Set some arguments to be keyword-only 
    Rationale: most args in ptable plotters are complicated, and set them as keyword-only might reduce ambiguity. 

### TODO

- [x] Update other ptable plotters (only those with new implementation)
- [x] Update README to include one example of this functionality


pymatviz/ptable.py
```diff
@@ -416,6 +422,10 @@ def add_child_plots(
             on_empty: Whether to "show" or "hide" tiles for elements without data.
         """
         for element in Element:
+            # Hide f-block
+            if self.hide_f_block and (element.is_lanthanoid or element.is_actinoid):
```
janosh, line 426: could also use [`is_rare_earth_metal`](https://github.com/materialsproject/pymatgen/blob/a9da16e86f072febe3f25efe73a38f9ac611acd1/pymatgen/core/periodic_table.py#L690) here

DanielYang59: Thanks!

DanielYang59: The definition from `pymatgen` is a bit weird. By definition, the [rare-earth element](https://en.wikipedia.org/wiki/Rare-earth_element) should be Lanthanides, yttrium (Y) and Scandium (Sc)?

janosh: i think there are [diverging opinions about this](https://www.google.com/search?q=are+actinides+rare+earth+metals&oq=are+actinides+rare). not sure `pymatgen` is the most widely held one but it would be a process to change it now

DanielYang59: I can see that now. But what about Y and Sc? This [doesn't seem controversial](https://www.google.com/search?q=are+Sc+and+Y+rare+earth+metals&sca_esv=e6402e358686dbdc&sxsrf=ADLYWIJsFZ1TajKcfwoF11R_eydLr4oZig%3A1715169303723&ei=F2g7ZoTnK7zvseMP1KS_8Ak&ved=0ahUKEwjEzs_I__2FAxW8d2wGHVTSD54Q4dUDCBA&uact=5&oq=are+Sc+and+Y+rare+earth+metals&gs_lp=Egxnd3Mtd2l6LXNlcnAiHmFyZSBTYyBhbmQgWSByYXJlIGVhcnRoIG1ldGFsczIIEAAYgAQYogQyCBAAGIAEGKIEMggQABiABBiiBDIIEAAYgAQYogQyCBAAGIAEGKIESLAPULYDWKQNcAJ4AZABAZgByAKgAY4MqgEHMC43LjAuMbgBA8gBAPgBAZgCBqAC2AXCAgoQABiwAxjWBBhHwgIIEAAYBxgIGB7CAggQABgFGAcYHpgDAIgGAZAGCpIHAzIuNKAHnhg&sclient=gws-wiz-serp).

janosh: don't know how the initial implementation was motivated and it would be involved to change it now. feel free to add a PR adding the links we shared above to the [is_rare_earth_metal](https://github.com/materialsproject/pymatgen/blob/a9da16e86f072febe3f25efe73a38f9ac611acd1/pymatgen/core/periodic_table.py#L690) doc string so users are at least aware that different definitions exist

DanielYang59: Sure. I would do it.

pymatviz/ptable.py
```diff
@@ -636,6 +651,7 @@ def ptable_heatmap(
     value_font_size: int = 12,
     tile_size: float | tuple[float, float] = 0.9,
     rare_earth_voffset: float = 0.5,
```
janosh, line 653: maybe rename `rare_earth_voffset` to `f_block_voffset` for consistency? don't imagine this kwarg is used a lot so could just make it a breaking rename. or use 

```py
if "rare_earth_voffset" in kwargs:
    f_block_voffset = kwargs.pop("rare_earth_voffset")
    print("deprecation warning: use f_block_voffset instead of rare_earth_voffset
```

during a migration period

janosh (changes requested):
this looks great! 👍 all that's missing are some unit tests

pymatviz/ptable.py
DanielYang59: Thanks for proposing this... Maybe a breaking change? As I don't assume plotters to be integrated into automatic workflows much?

pymatviz/ptable.py
```diff
@@ -793,6 +796,9 @@ def tick_fmt(val: float, _pos: int) -> str:
     ) | (text_style or {})
 
     for symbol, row, column, *_ in df_ptable.itertuples():
+        if hide_f_block and (row in (6, 7)):
```
DanielYang59, line 799: I thought I would do this after refactor all other plotters with the new `PTableProjector`, but thanks for fixing this.

DanielYang59:
> this looks great! 👍 

Thanks for reviewing. 

> all that's missing are some unit tests

Exactly, but I found it tricky to write unit tests for plotters. Do you have any suggestions? 

Assert a `Figure` being generated doesn't really check the results of a plotter. And I have to visually inspect the result (this sounds quite hard to be automated, unlike data processors).

janosh:
> Assert a Figure being generated doesn't really check the results of a plotter. And I have to visually inspect the result (this sounds quite hard to be automated, unlike data processors).

i would test that `if hide_f_block is True`, the plot contains 30 fewer element tiles and maybe confirm that all the lanthanide and actinide are absent in `labels = {txt.get_text() for txt in ax.texts}`

DanielYang59:
Sounds great. I would add this later :)

janosh:
> Sounds great. I would add this later :)

great. i think once this is merged, we're ready for a v0.8.2 release so people who install from pypi can start using `ptable_heatmap_splits`.

pymatviz/ptable.py
```diff
@@ -332,17 +334,23 @@ def __init__(
...
+
         # Preprocess data
         self.data: pd.DataFrame = data
 
         # Initialize periodic table canvas
         n_periods = df_ptable.row.max()
         n_groups = df_ptable.column.max()
 
+        if self.hide_f_block:
+            n_periods -= 3
```
DanielYang59, line 352: I removed the bottom three rows `if hide_f_block`, which not only enables unit test check for `len(axes)`, but also helped avoid the extra space at the bottom. 

@janosh. Do you have any further comments? Thanks!

janosh: > just testing one of them should suffice.

i'd prefer to test multiple here

assets/ptable-heatmap-splits.svg
janosh, line 1: nice, much better now. didn't actually notice all the white space before

janosh (changes requested):
i think we can be a bit smarter about the `hide_f_block` default behavior. let's set it to `None` by default and then use `True` or `False` depending on whether we have non-zero values for any rare earths. passing `True` or `False` explicitly of course will override this behavior

DanielYang59:
> i think we can be a bit smarter about the `hide_f_block` default behavior. let's set it to `None` by default and then use `True` or `False` depending on whether we have non-zero values for any rare earths. passing `True` or `False` explicitly of course will override this behavior

Excellent idea. Let's do it.

tests/test_ptable.py
```diff
@@ -503,29 +510,35 @@ def test_ptable_hists(
...
     fig = ptable_scatters(
         data={
             "Fe": [[1, 2, 3], [4, 5, 6]],
             "O": [[10, 11], [12, 13]],
-        }
+        },
+        hide_f_block=hide_f_block,
     )
     assert isinstance(fig, plt.Figure)
+    assert len(fig.axes) == 126 if hide_f_block else 181
```
DanielYang59, line 538: Not quiet sure why the number of axes is slightly different (127 for heatmap plotters).

pymatviz/ptable.py
```diff
@@ -332,17 +334,35 @@ def __init__(
...
         self.data: pd.DataFrame = data
 
+        if hide_f_block is None:
+            hide_f_block = bool(
+                {
+                    atom_num
+                    for atom_num in [*range(57, 72), *range(89, 104)]  # rare earths
+                    # check if data is present for f-block elements
+                    if (elem := Element.from_Z(atom_num).symbol) in self.data.index  # type: ignore[union-attr]
+                    and self.data.loc[elem, "Value"]  # type: ignore[union-attr]
```
janosh, line 353: one more thing to change here: it's always a good to maintain [SSOT](https://en.wikipedia.org/wiki/Single_source_of_truth) to avoid drift and so it's to change things in the future. the `"Value"` is hard-coded in many places in this file. would be better to add a [new key in init.py](https://github.com/danielyang59/pymatviz/blob/69ea94920082b814d682ed3b3ccda5999b018009/pymatviz/__init__.py#L91)

DanielYang59: Yes, as we're hard code the column names: https://github.com/janosh/pymatviz/blob/39fd71e5e105574ed1e8d1e3174f9ae4257f95be/pymatviz/ptable.py#L217-L254

Maybe it's good to define a standard name of both columns in case we need to change them in the future?

janosh (approved):
thanks so much for this @DanielYang59! 👍

pymatviz/ptable.py
```diff
@@ -235,8 +239,8 @@ def data_preprocessor(data: SupportedDataType) -> pd.DataFrame:
         OR
         >>> data_df: pd.DataFrame = pd.DataFrame(
             data_dict.items(),
-            columns=["Element", "Value"]
-            ).set_index("Element")
+            columns=[{element_col}, {value_col}]
```
DanielYang59, line 242: It looks like this breaks the site builder workflow. I should have used a raw string instead. Please help me fix this. Thanks!

DanielYang59:
Great to collaborate with you on this one. Thanks!

pymatviz/ptable.py
janosh: will fix, PR incoming for SSOT of keys across the whole code base

PR #143, opened 2024-05-11
Add `pymatviz/enums.py` for SSOT on dataframe column and dict key names

janosh:
prompted by https://github.com/janosh/pymatviz/pull/140#discussion_r1597454263


janosh:
@DanielYang59 curious to hear your thoughts on this? is this good or overkill?

DanielYang59:
I absolutely love this. Maybe make `enum.py` private? I think user do not need to access it?

janosh:
> I absolutely love this. Maybe make enum.py private? I think user do not need to access it?

thanks! i think it doesn't hurt to help users follow `pymatviz` terminology if they choose to. consistency helps interoperability

PR #146, opened 2024-05-15
Fix `ptable_heatmap_splits` `TypeErrors`

### LINKED ISSUE #145: `ptable_heatmap_splits` raises unexpected `TypeErrors`
@DanielYang59 i noticed a few bugs in `ptable_heatmap_splits`

```py
ptable_heatmap_splits({"H": [1, 2]})
>>> TypeError: 'NoneType' object is not callable
```

you need to explicitly pass a `colormap` to avoid this. we should set a default. also:
```py
ptable_heatmap_splits({"H": {"A": 1, "B": 2}.values()}, colormap="RdBu_r")
>>> TypeError: Unsupported data type.
```

we currently don't handle type [`dict_values`](https://docs.python.org/3/c-api/dict.html#c.PyDict_Values). if you wrap `dict_values` in `list`, it works:

```py
ptable_heatmap_splits({"H": list({"A": 1, "B": 2}.values())}, colormap="RdBu_r")
```

janosh:
closes #145

- Use `viridis` as default colormap for ptable plotters
- Handle `dict_values` and update unit test


DanielYang59:
It looks like `ReprEnum` was [added at Python 3.11](https://docs.python.org/3/library/enum.html#enum.ReprEnum)

DanielYang59:
@janosh Please review this. Thanks!

tests/test_ptable.py
```diff
@@ -276,6 +276,13 @@ def test_ptable_heatmap_splits(hide_f_block: bool) -> None:
         for elem in Element
     }
 
+    # Also test different data types
+    data_dict["H"] = {"a": 1, "b": 2}.values()
+    data_dict["He"] = [1, 2]
+    data_dict["Li"] = np.array([1, 2])
+    data_dict["Be"] = 1
+    data_dict["B"] = 2.0
```
janosh, line 284: thanks, this is good! but i think these lines should go in `TestDataPreprocessor`, no?

DanielYang59: `TestDataPreprocessor` already has these test cases, so I assume it might be good to duplicate these in a real plotter :)

janosh: true, we can have them in both. it doesn't have the `dict_values` yet. i'll add those

https://github.com/janosh/pymatviz/blob/d391350c79a46149f6a8a5d8059e33dc7752aff7/tests/test_ptable.py#L41

DanielYang59: Was thinking the same, so 25f8d708782ec7cc6ac2a7c6db3f8aeb856b0b59 :)

PR #149, opened 2024-05-23
[Enhancement/Breaking] Refactor `ptable_hists` 

### LINKED ISSUE #142: [Enhancement] Refactor`ptable_hists` and `ptable_heatmap` with new `PTableProjector`
### Refactor `ptable_hists` and `ptable_heatmap` with new `PTableProjector`.

### LINKED ISSUE #151: [Enhancement] Allow adding `elem_type_legend` and color element tiles by `color_elem_strategy` for ptable scatter/line plotters
[Enhancement] Allow adding `elem_type_legend` and color element tiles by `color_elem_strategy` for ptable scatter/line plotters

DanielYang59:
## Summary
- [Enhancement] Refactor `ptable_hists` with new implementation, partly close #142 
- [Enhancement] Allow background/symbol coloring based on element type for scatter/line plotters, to close #151

### Breaking changes
- [**Fix**] Rename `child_args` to `child_kwargs`
- Force all args other than `data` in ptable plotters to be keyword-only in cb643ccc269938cbb9a630763b6e18b60cea018b
- Rename `*_kwds` to `*_kwargs` for consistency with [PEP docs](https://peps.python.org/pep-0468/) (with [a few exceptions in PEP as well](https://stackoverflow.com/questions/6435319/what-is-the-python-convention-kwargs-vs-kwds-vs-kw))
- `data` as `pd.DataFrame(some_dict)` isn't supported anymore, users need to pass the `some_dict` directly. For example: https://github.com/janosh/pymatviz/blob/5dd53ef31e3f1ef13aede234ef379d2eb6966e69/tests/test_ptable.py#L588-L595

#### Adjust args ordering in ptable plotters

Currently the args ordering is adjusted to the following ordering (now we're forcing them to be keyword-only):
- Function-specific args come first, e.g. `bins` for histograms, `start_angle` for `heatmap_splits`
- Global settings come before local settings, e.g. `hide_f_block/plot_kwargs` come before `ax/symbol/cbar`

<details>
  <summary>Example</summary>

    def ptable_heatmap_splits(
    data: pd.DataFrame | pd.Series | dict[str, list[list[float]]],
    *,
    ##### Function-specific #####
    start_angle: float = 135,

    ##### Figure-scope settings #####
    colormap: str | Colormap = "viridis",
    hide_f_block: bool | None = None,
    plot_kwargs: dict[str, Any]
    | Callable[[Sequence[float]], dict[str, Any]]
    | None = None,

    ##### Axis-scope settings #####
    ax_kwargs: dict[str, Any] | None = None,

    ##### Local object: symbol #####
    symbol_pos: tuple[float, float] = (0.5, 0.5),

    ##### Local object: cbar #####
    cbar_coords: tuple[float, float, float, float] = (0.18, 0.8, 0.42, 0.02),
    cbar_title: str = "Values",
    cbar_title_kwargs: dict[str, Any] | None = None,
    ) -> plt.Figure:

</details>

### Update `PTableProjector`
- Consumed `add_element_type_legend` function as a method
- Multiple useful properties: `hide_f_block`, `elem_type_colors`, `elem_colors` and `hide_f_blocks` with setter extracted.


DanielYang59:
### `color_elem_types`
https://github.com/janosh/pymatviz/blob/5dd53ef31e3f1ef13aede234ef379d2eb6966e69/pymatviz/ptable.py#L1422-L1424

We might separate `color_elem_types` into two args: 
- One `strategy: Literal["symbol", "background", "both", "off"]` to select the coloring strategy.
- Another optional `elem_type_colors: dict[str, str] | None` to overwrite default element type colors if desired.

Otherwise, users cannot overwrite the colors and select strategy at the same time (i.e. if they pass a `dict`, they then cannot select from "symbol"/"background" or "both").


### Function `add_element_type_legend`

Might be good to completely consume `add_element_type_legend` into `PTableProjector`, as another method to handle legend.

janosh:
> We might separate `color_elem_types` into two args:
> 
> * One `strategy: Literal["symbol", "background", "both", "off"]` to select the coloring strategy.
> * Another optional `elem_type_colors: dict[str, str] | None = None` to overwrite default element type colors if desired.
> 
> Otherwise, users cannot overwrite the colors and select strategy at the same time (i.e. if they pass a `dict`, they then cannot select a strategy from "symbol", "background" or "both").

that sounds sensible! feel free to make the change. we should also list valid keys for the element types dict in the doc string. something i forgot.
i think i prefer `elem_type_color_strategy` over `strategy` to be more descriptive.

DanielYang59:
> i think i prefer `elem_type_color_strategy` over `strategy` to be more descriptive.

That's exactly what I used in 5d435895149622859140eb1b05ad3137cd835b73 :). I wrote a shorthand in the comments and sorry for the confusion

tests/test_ptable.py
```diff
@@ -506,39 +555,47 @@ def test_ptable_heatmap_plotly_label_map(
 
 
 @pytest.mark.parametrize(
-    "data, symbol_pos, anno_kwds, hist_kwds",
```
DanielYang59, line 509: Need confirmation on this change.

Previously the `ptable_hists` has both `symbol_kwargs` and `anno_kwds`. In the new implementation, support for `annotate` is completely removed.

I'm not sure if we need the `annotate` in the first place, because it appears we just need a `symbol` to show for each element tile (users could tweak the text from `symbol_text`), we don't seem to need another annotation, which might be too small to show properly anyway. 

Can you please suggest on this?@janosh

janosh: it works for me locally. maybe try updating your `matplotlib`? i restored the `mpl.colors.to_rgb` part in order to apply an alpha channel. i think background colors need to be applied as a light touch only, otherwise the plot becomes unreadable.
image

**without alpha channel**:

72b8caff-2ee6-4651-bd30-4797ea4d5a3c

DanielYang59: Yes this `alpha` looks indeed a must. Don't know about why, would have another try tomorrow. Thanks!

DanielYang59: Update on this one, `plt.show()` seems to work fine, but `save_and_compress_svg` saved the figure without background color. I'm looking into it.

pymatviz/ptable.py
```diff
@@ -428,12 +429,8 @@ def get_elem_type_color(
         default: str = "white",
     ) -> str:
         """Get element type based color."""
-        try:
-            elem_type = df_ptable.loc[elem_symbol, "type"]
-            return self.elem_type_colors.get(elem_type, default)
-
-        except KeyError:
```
DanielYang59: This was intended to guard against a non-existent element (happy to discuss this if you think this is unnecessary), I would revert this and issue a warning in such case.

```python
    def test_get_elem_type_color(self) -> None:
        projector = PTableProjector(data=self.test_dict)

        assert projector.get_elem_type_color("H") == "green"
        assert projector.get_elem_type_color("Fe") == "blue"
        assert projector.get_elem_type_color("NONE", "Hello") == "Hello"
```

DanielYang59: I decide to remove this as we cannot expect the code to work for a non-existent element. This doesn't look necessary.

DanielYang59:
Should be ready for review @janosh. Thanks for your time :)

janosh:
What's the reason for dropping support for

> data as pd.DataFrame(some_dict) isn't supported anymore, users need to pass the some_dict directly.

DanielYang59:
> What's the reason for dropping support for
> 
> > data as pd.DataFrame(some_dict) isn't supported anymore, users need to pass the some_dict directly.

Thanks for asking. 

Currently `data` as `dict` is preprocessed internally: https://github.com/janosh/pymatviz/blob/5dd53ef31e3f1ef13aede234ef379d2eb6966e69/pymatviz/ptable.py#L266-L269

So I should not say support for `pd.DataFrame(dict)` is completely dropped, but user need to preprocess the same way themselves. Might be easier for them to pass `dict` directly as `data`?

janosh:
It's nice to maximize input flexibility. Ideally, any data format with a unique semantic meaning (atomic numbers or element symbols as dict keys, or as dataframe index or even a list with less than 118 elements) should be supported.

DanielYang59:
I think it makes sense to support such "transposed" `pd.DataFrame`, I would implement this tomorrow.

tests/test_ptable.py
```diff
@@ -49,7 +45,7 @@ class TestDataPreprocessor:
         "Li": np.array([6.0, 8.0]),  # float array
         "Na": 11.0,  # float
         "Mg": {"a": -1, "b": 14.0}.values(),  # dict_values
-        "Al": {-1, 2.3},  # mixed int/float set
```
DanielYang59, line 52: I don't think user should use unordered data for plotting, this is not supported by `pd.DataFrame` constructor:
```python
  File "/Users/yang/developer/pymatviz/test_df.preprocess.py", line 18, in <module>
    input_df_0: pd.DataFrame = pd.DataFrame(test_dict)
                               ^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/yang/developer/pymatviz/venv/lib/python3.12/site-packages/pandas/core/frame.py", line 778, in __init__
    mgr = dict_to_mgr(data, index, columns, dtype=dtype, copy=copy, typ=manager)
          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/yang/developer/pymatviz/venv/lib/python3.12/site-packages/pandas/core/internals/construction.py", line 503, in dict_to_mgr
    return arrays_to_mgr(arrays, columns, index, dtype=dtype, typ=typ, consolidate=copy)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/yang/developer/pymatviz/venv/lib/python3.12/site-packages/pandas/core/internals/construction.py", line 119, in arrays_to_mgr
    arrays, refs = _homogenize(arrays, index, dtype)
                   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/yang/developer/pymatviz/venv/lib/python3.12/site-packages/pandas/core/internals/construction.py", line 629, in _homogenize
    val = sanitize_array(val, index, dtype=dtype, copy=False)
          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/yang/developer/pymatviz/venv/lib/python3.12/site-packages/pandas/core/construction.py", line 642, in sanitize_array
    _sanitize_non_ordered(data)
  File "/Users/yang/developer/pymatviz/venv/lib/python3.12/site-packages/pandas/core/construction.py", line 693, in _sanitize_non_ordered
    raise TypeError(f"'{type(data).__name__}' type is unordered")
TypeError: 'set' type is unordered
```

DanielYang59:
@janosh. I just updated the data processor trying out best to fix `pd.DataFrame` that doesn't come in expected format(missing column/row names or transposed). So any `df` should work as long as there is a column/row containing element names, and other columns are all floats/ints.

There is a pitfall though, but I don't think it's something our code should handle:
```python
test_dict = {
        "H": 1,  # int
        "He": [2.0, 4.0],  # float list
        "Li": np.array([6.0, 8.0]),  # float array
        "Na": 11.0,  # float
    }

>>> pd.DataFrame(test_dict)
   H   He   Li    Na
0  1  2.0  6.0  11.0
1  1  4.0  8.0  11.0
```

`pandas` would pad `df` to the same length, but I don't think this is something we should fix . As from our view, the `df` should be all we got, so it's the user's responsibility to use it as properly as possible, right?

tests/test_ptable.py
janosh: unordered data is fine for a histogram, no? i think in this case, `DataFrame` has stricter requirements than we need to enforce

janosh:
> As from our view, the df should be all we got, so it's the user's responsibility to use it as properly as possible, right?

agreed. 👍

tests/test_ptable.py
DanielYang59: Yes it's fine for histogram, but this `TestDataPreprocessor` is for more than histogram.

DanielYang59: Maybe I wasn't make myself clear, the problem is not about our data processor, pandas cannot construct a `df` from a dict with `set` as values:
```python
test_dict = {
        "H": 1,  # int
        "He": {2.0, 4.0},  # float list
    }

input_df_0: pd.DataFrame = pd.DataFrame(test_dict)  # TypeError: 'set' type is unordered
```

janosh: No that part I understood :)

DanielYang59:
@janosh Do you have further comments or can we merge this one?

After this I might go on to re-enable 3rd color dimension for `ptable_scatters` plotter, and then get my hands on refactoring the `ptable_heatmap` plotter.

There seems to be quite a few internal label/text formatter inside the `ptable_heatmap`, so I guess it would be quite some work too. Also it is depended on by `ptable_heatmap_ratio`, hopefully we could get them done together :)

pymatviz/ptable.py
```diff
@@ -522,16 +619,71 @@ def add_colorbar(
...
 
         # Set colorbar title
-        title_kwargs = title_kwargs or {}
-        title_kwargs.setdefault("fontsize", 12)
-        title_kwargs.setdefault("pad", 10)
-        title_kwargs["label"] = title
-
         cbar_ax.set_title(**title_kwargs)
 
+    def add_elem_type_legend(
```
janosh, line 620: i think something got lost here. the colors no longer default to `ELEM_CLASS_COLORS`?

pymatviz/ptable.py
```diff
@@ -255,9 +193,77 @@ def data_preprocessor(data: SupportedDataType) -> pd.DataFrame:
...
+            return fixed_df
+
+        # Try to search for elements as a row
+        fixed_df = fix_df_elem_as_col(df.transpose())
+        if fixed_df is not None:
+            return fixed_df
+
+        raise RuntimeError("Cannot fix provided DataFrame.")
+
+    def handle_missing_and_anomaly(
```
janosh, line 249: what's left to be done here?

janosh (changes requested):

pymatviz/ptable.py
DanielYang59: It still is, I packed everything inside the following property:
```python
    @elem_type_colors.setter
    def elem_type_colors(self, elem_type_colors: dict[str, str] | None) -> None:
        default_elem_type_colors: dict[str, str] = {
            "Diatomic Nonmetal": "green",
            "Noble Gas": "purple",
            "Alkali Metal": "red",
            "Alkaline Earth Metal": "orange",
            "Metalloid": "darkgreen",
            "Polyatomic Nonmetal": "teal",
            "Transition Metal": "blue",
            "Post Transition Metal": "cyan",
            "Lanthanide": "brown",
            "Actinide": "gray",
            "Nonmetal": "green",
            "Halogen": "teal",
            "Metal": "lightblue",
            "Alkaline Metal": "magenta",
            "Transactinide": "olive",
        }

        self._elem_type_colors = default_elem_type_colors | (elem_type_colors or {})
```

pymatviz/ptable.py
DanielYang59: Good catch, it was not finished when we first refactor the ptable plotter with `PTableProjector`, and I have been procrastinating on this :( But I would do it soon :)

We would need to implement infinity and `NaN` handling here, and also need to update the unit test for this.

janosh: we don't need to tackle it in this PR. let me know if you prefer to do this later

pymatviz/ptable.py
```diff
@@ -599,7 +599,9 @@ def add_colorbar(
         """
         # Update colorbar kwargs
         cbar_kwargs = {"orientation": "horizontal"} | (cbar_kwargs or {})
-        title_kwargs = {"fontsize": 12, "pad": 10} | (title_kwargs or {})
+        title_kwargs = {"fontsize": 12, "pad": 10, "label": title} | (
```
janosh, line 602: oops, had `"label": title` in there initially but then accidentally pressed undo once too many 😅

DanielYang59: All good. Typos happen :)

pymatviz/ptable.py
DanielYang59: Well... Let's handle this later then, for two reasons:
One, this PR is already pretty huge.

More importantly, I current don't have a solid (elegant) solution in my mind, on how to maintain the shape of nested arrays. If we were to fix bad values in the provided `DataFrame`, we would need to:
- First, explode (flatten) the `DataFrame` to check for NaN or infinity.
- If found, we would need to replace such values, but this seems a bit tricky as our supported data types are too flexible (and unpredictable).

For data without nested structure (`ptable_hists` and `ptable_heatmap/heatmap_splits`), this is fairly easy, we could simply record its original shape, and recover it:https://github.com/janosh/pymatviz/blob/5dd53ef31e3f1ef13aede234ef379d2eb6966e69/tests/test_ptable.py#L46-L53

But for plotters with deeper nested structure, this sounds a bit messy, for example `ptable_scatters`: https://github.com/janosh/pymatviz/blob/5dd53ef31e3f1ef13aede234ef379d2eb6966e69/tests/test_ptable.py#L576-L580

We could certain record its shape recursively, but I'm hoping that we could have a cleaner solution. Also during such handling, we also need to group data (i.e. we have to separate data for `x`/`y` and so on, instead of using the global average to infer a missing data for `y`).

tests/test_ptable.py
janosh: oh, that's good to know. maybe we should change the default `save_and_compress_svg(transparent=False)`?

janosh (approved):
thanks so much for these improvements. i'm quite optimistic now that we can standardize data preprocessing across `ptable` plotter functions which would be huge! will make this module much more maintainable going forward

pymatviz/utils.py
```diff
@@ -46,7 +46,7 @@
 elements_csv = f"{ROOT}/pymatviz/elements.csv"
 df_ptable = pd.read_csv(elements_csv, comment="#").set_index("symbol")
 
-ELEM_CLASS_COLORS: Final = {
```
DanielYang59, line 49: Thanks! I was not aware its existence here. 

And yes I renamed everything to "element type" instead of "element class" (also used "elem" over "ele" as the abbreviation), because previously these two are used inconsistently across ptable plotters :)

DanielYang59:
> thanks so much for these improvements. 

Thanks for saying this. I'm also really, really and really happy we manage to get rid of the huge amount of repeated code across ptable plotter :) I like current code style so much more :) Everything looks so modular 

> i'm quite optimistic now that we can standardize data preprocessing across `ptable` plotter functions which would be huge! will make this module much more maintainable going forward

Yes we are getting very close I believe, perhaps the previously mentioned `handle_missing_and_anomaly` is the only thing left? What else is missing now? 

And I'm thinking about the possibility to share data preprocessing for other modules. ptable plotter seems to be the most complex one with a set of data for each element, for other plotters there is usually a set of data (a set of `x`/`y`). So I guess we just need to write a simple data processor for other plotters, and reuse it in ptable plotters?

DanielYang59:
Meanwhile, is the `pymatviz` logo missing from your side? https://github.com/janosh/pymatviz/pull/149#issuecomment-2130683734

DanielYang59:
Weird. So let it be.

It seems ` is not allowed for docstrings? The CI test failed https://github.com/janosh/pymatviz/actions/runs/9251916398/job/25448427714

DanielYang59:
> you probably know better than me. 

:) Let's finish `handle_missing_and_anomaly` then, I would consider this an edge case (why would people feed NaN or inifinity into a plotter!?), but it's still good to handle it :)

> but i was thinking of the 3 or 4 TODOs you left in the code `TODO: how to control this from external?`
> 
> https://github.com/janosh/pymatviz/blob/ff600009f9eab4a522c3cfbf0cf8afc2b8bd5654/pymatviz/ptable.py#L811-L816

Yep, I got caught by you again. It's mostly related to the style of ticks, currently the tick style is hard-coded and cannot be modified from the child plotter's side. It's certain better to enable such fine tunes if we want to provide publication-quality figures. But I don't know how to achieve this now.

janosh:
> It seems ` is not allowed for docstrings? The CI test failed https://github.com/janosh/pymatviz/actions/runs/9251916398/job/25448427714

i need to do completely revamp the docs at some point. the current setup is way too brittle and places constraints on what you can write in python doc strings which is very annoying (due to passing the doc strings through mdsvex which is a whole template language that gives special meaning to certain characters, esp. to curly braces which open an Svelte inline code evaluation context)

janosh:
> Let's finish handle_missing_and_anomaly then, I would consider this an edge case (why would people feed NaN or inifinity into a plotter!?), but it's still good to handle it :)

i agree we don't need to handle every conceivable edge cases. but it would be nice to detect edge cases we don't handle and issue helpful error messages in those cases.

after that, i think we can release v0.8.3 🚀

DanielYang59:
> > It seems ` is not allowed for docstrings? The CI test failed https://github.com/janosh/pymatviz/actions/runs/9251916398/job/25448427714
> 
> i need to do completely revamp the docs at some point. the current setup is way too brittle and places constraints on what you can write in python doc strings which is very annoying (due to passing the doc strings through mdsvex which is a whole template language that gives special meaning to certain characters, esp. to curly braces which open an Svelte inline code evaluation context)

Feel free to ping me if I could help :)



> > Let's finish handle_missing_and_anomaly then, I would consider this an edge case (why would people feed NaN or inifinity into a plotter!?), but it's still good to handle it :)
> 
> i agree we don't need to handle every conceivable edge cases. but it would be nice to detect edge cases we don't handle and issue helpful error messages in those cases.

Yes I'm already getting my hands on this.

> after that, i think we can release v0.8.3 🚀

Exciting! I would try to fix the 3rd color dimension for `ptable_scatters` soon :) After this, let's march to refactor the `ptable_heatmap`

PR #152, opened 2024-05-27
Handle missing value (NaN) and infinity for ptable data

DanielYang59:
## Summary
- Separate data processing as a separate module
- Handle missing value (NaN) and infinity for ptable data
- Pack kwargs to `ax.tick_params()` call for easier manipulation from outside

## Separate data processing as a separate module

As our data preprocessing function gets larger (and more powerful), might be good to put it in a separate module. I didn't do this when first implemented because I thought it is specific to ptable plotter, but as we're pushing towards better data preprocessing across the code base, it might be the right time to do this.

Also I added a new module instead of putting it into `utils.py`, because I thought this would be more self-descriptive of its functionality.

## Handle missing value (NaN) and infinity for ptable data

- [x] `check_for_missing_inf` to check if there is missing value or infinity.
- [x] `get_df_nested_level` to get max nested level to decide if we can fix or not.
- [x] Add missing/infinity imputation function.
- [x] Clarify example in `preprocess_ptable_data` docstring.
- [x] [**Need help**] Fix `mymy` error for `child_plotter` call with `tick_kwargs`.


pymatviz/ptable.py
```diff
@@ -503,7 +355,7 @@ def add_child_plots(
                 continue
 
             # Call child plotter
-            child_plotter(ax, plot_data, **child_kwargs)
+            child_plotter(ax, plot_data, tick_kwargs=tick_kwargs, **child_kwargs)  # type: ignore[call-arg]
```
DanielYang59, line 358: Not sure why `mypy` reported:
```python
pymatviz/ptable.py:358: error: Unexpected keyword argument "tick_kwargs"  [call-arg]
```

DanielYang59:
@janosh  Can you please review this?

I'm not sure why this `mypy` failure though: https://github.com/janosh/pymatviz/pull/152#discussion_r1617081403

pymatviz/ptable.py
janosh: [have a look at the docs for type-hinting `Callable`](https://docs.python.org/3/library/typing.html#annotating-callable-objects). it's quite restricted on what signatures it can model. to be maximally permissive in the inputs while still constraining the output, you can use `Callable[..., SomeReturnType]`

pymatviz/_data.py
janosh, line 1: do we really want this file to be marked private? also maybe `preprocess_data.py` or `normalize_data.py` would be a better name? no reason to shy away from long but descriptive names 😄

DanielYang59: > do we really want this file to be marked private? 

I would personally prefer to make it private because users normally would not need to access this module, this is kind of self-descriptive I guess?

> also maybe `preprocess_data.py` or `normalize_data.py` would be a better name? no reason to shy away from long but descriptive names 😄

This is a great point!

pymatviz/ptable.py
DanielYang59: Beautiful. That's a wonderful fix!

pymatviz/_data.py
```diff
@@ -0,0 +1,303 @@
...
+        col (str): Name of the column to check.
+        max_level (int): Max nest level to try up to. Defaults to 10.
+
+    Returns:
+        int: The maximum nest level.
+    """
+    current_level = 0
+    while current_level <= max_level:
+        # Check if any element in the column is a list
+        if df_in[col].map(lambda x: isinstance(x, list)).any():
```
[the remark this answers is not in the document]
DanielYang59: > what if the column values are arrays instead of lists? 

Great! I already fixed this and updated the unit test.

> also, maybe instead of looping it would be better to call `np.array()` on one of the lists and then return `arr.ndim`

This sounds like a wonderful solution. But I'm still trying to get it to work, because we might be expecting different dimensions for different elements

```python3
df_level_2 = pd.DataFrame(
    {
        "Fe": [1, 2, 3],
        "O": [[4, 5], [6, 7]],  # get max level
    }.items(),
    columns=[Key.element, Key.heat_val],
).set_index(Key.element)

column_values = df_level_2[Key.heat_val].tolist()
array = np.array(column_values)
```

Which raises:
```python3
  File "/Users/yang/developer/pymatviz/test_df.py", line 17, in <module>
    array = np.array(column_values)
            ^^^^^^^^^^^^^^^^^^^^^^^
ValueError: setting an array element with a sequence. The requested array has an inhomogeneous shape after 1 dimensions. The detected shape was (2,) + inhomogeneous part.
```

pymatviz/_preprocess_data.py
```diff
@@ -69,22 +69,11 @@ def get_df_nest_level(df_in: pd.DataFrame, col: str, max_level: int = 10) -> int
...
-        # Check if any element in the column is a list or np.ndarray
-        if df_in[col].map(lambda x: isinstance(x, (list, np.ndarray))).any():
-            current_level += 1
-            # Explode the lists into separate rows
-            df_in = df_in.explode(col)
-        else:
-            return current_level
-
-    return current_level
+    return df_in[col].apply(lambda val: np.array(val).ndim).max()
```
janosh, line 76: [`np.asarray`](https://numpy.org/doc/stable/reference/generated/numpy.asarray.html#numpy.asarray) has the advantage of creating a new array only when needed, i.e. only when it's not already an array and could speed this function up on large dataframes

janosh: also, given it's a one-liner, do we need to make it a function? it's only used once, right?

janosh (changes requested):

pymatviz/_preprocess_data.py
janosh: btw, `df.map` is faster than `df.apply` so i always use the former unless i actually need `apply` (basically never)

pymatviz/_preprocess_data.py
```diff
@@ -0,0 +1,292 @@
...
+        return np.array(
+            [
+                replacement_nan
+                if pd.isna(val)
+                else replacement_inf_pos
+                if val == np.inf
+                else replacement_inf_neg
+                if val == -np.inf
+                else val
+                for val in value
```
janosh, line 138: using a dict here would probably be more readable:

```
{pd.na: replacement_nan, np.inf: replacement_inf_pos, -np.inf: replacement_inf_neg}.get(val, val) for ...
```

pymatviz/_preprocess_data.py
```diff
@@ -227,7 +227,7 @@ def fix_df_elem_as_col(df: pd.DataFrame) -> pd.DataFrame:
             value_cols = [
                 col_name for col_name in new_df.columns if col_name != Key.element
             ]
-            new_df[Key.heat_val] = new_df[value_cols].map(list)
```
DanielYang59: Not sure why `map` for this? Seem to fail for `int`

pymatviz/_preprocess_data.py
```diff
@@ -0,0 +1,264 @@
...
+        fixed_df = fix_df_elem_as_col(df)
+        if fixed_df is not None:
+            return fixed_df
+
+        # Try to search for elements as a row
+        fixed_df = fix_df_elem_as_col(df.transpose())
+        if fixed_df is not None:
+            return fixed_df
+
+        raise ValueError(f"Cannot normalize the given DataFrame:\n{df.describe()}")
```
janosh, line 224: can we add a test case that triggers this error msg?

pymatviz/_preprocess_data.py
```diff
@@ -0,0 +1,264 @@
...
+        else np.array([replacements.get(v, v) for v in val])
+    )
+
+    return df_in
+
+
+def preprocess_ptable_data(data: SupportedDataType) -> pd.DataFrame:
+    """Preprocess input data for ptable plotters, including:
+        - Convert all data types to pd.DataFrame.
+        - Impute missing values.
```
janosh, line 133: can you add one more sentence here to explain what actual steps are taken during data imputation?

pymatviz/_preprocess_data.py
DanielYang59: Done. Basically any `dataframe` that doesn't have a complete row/column of elements should trigger this.

pymatviz/_preprocess_data.py
DanielYang59: Done! It seems the `missing_strategy` arg for `preprocess_ptable_data` is forgotten :( Glad we caught this.

janosh:
almost done here! 🚀 just 1 more thing: i removed the normalization of dataframes that have a row of element symbols in [e3a26e5](https://github.com/janosh/pymatviz/pull/152/commits/e3a26e55259601be21289926def23f0d9e47db3c). feel free to push back if you think that's important but seemed more like an edge case to me and allowed to simplify the `preprocess_ptable_data` helper quite a lot

DanielYang59:
>  i removed the normalization of dataframes that have a row of element symbols in [e3a26e5](https://github.com/janosh/pymatviz/pull/152/commits/e3a26e55259601be21289926def23f0d9e47db3c). feel free to push back if you think that's important but seemed more like an edge case to me and allowed to simplify the `preprocess_ptable_data` helper quite a lot

Yes I was about to say this. The function you removed seems to be the one of the core improvement (and perhaps [our reason for starting this PR](https://github.com/janosh/pymatviz/pull/149#issuecomment-2131209589) and https://github.com/janosh/pymatviz/pull/149#issuecomment-2131225374), i.e. trying to correct the format of input `DataFrame` as long as there is a single column/row fully consisted of elements.


Without this, `DataFrame` need to strictly follow the requirement (index as elements, and a value column), and thus the unit test failed for the following case (with and without `transpose()`):
```python
test_dict = {
    "He": [2.0, 4.0],  # float list
    "Li": np.array([6.0, 8.0]),  # float array
    "Mg": {"a": -1, "b": 14.0}.values(),  # dict_values
}

# Elements as a column, and no proper row/column names
df_in = pd.DataFrame(test_dict).transpose()
df_out = preprocess_ptable_data(df_in)
```

pymatviz/_preprocess_data.py
```diff
@@ -0,0 +1,226 @@
...
+            try:
+                data_df = pd.melt(
+                    data.reset_index(),
+                    id_vars=data.index.names,
+                    var_name=Key.element,
+                    value_name=Key.heat_val,
+                ).set_index(Key.element)
+            except KeyError as exc:
+                raise KeyError(
+                    f"Cannot normalize the given DataFrame:\n{data!s}"
```
DanielYang59, line 193: I'm a bit concerned with the wording "normalize" here, I understand in this context it means the same as "re-format the DataFrame", but I assume it could be confused with the other meaning "to rescale data to a certain range", especially in the context of "data preprocess"

pymatviz/_preprocess_data.py
```diff
@@ -177,13 +178,12 @@ def format_pd_dataframe(df: pd.DataFrame) -> pd.DataFrame:
...
             # Copy and reset index to move element names to a column
-            new_df = df.copy()
-            new_df = new_df.reset_index()
+            new_df = df.copy().reset_index()
 
             # Find the column with element names
             elem_col_name = None
             for col in new_df.columns:
-                if set(new_df[col]).issubset(set(map(str, Element))):
+                if set(new_df[col]).issubset({str, Element}):
```
DanielYang59: It appears there is some misunderstanding for this, this tries to find a column fully consists of elements(`{"H", "He", "O"}` for instance) by checking if it's a subset of all elements (`set(map(str, Element))`), it doesn't check the type of the values.

pymatviz/_preprocess_data.py
janosh: good point. how about "normalize" -> "handle"? maybe we should also add a sentence what needs to change for the dataframe to become valid

pymatviz/_preprocess_data.py
janosh: oh, now i get it. you're right, i misinterpreted this

janosh (approved):
thanks @DanielYang59 👍 and sorry for some of the extra work i caused 😅

pymatviz/_preprocess_data.py
DanielYang59: > good point. how about "normalize" -> "handle"? 

I would personally prefer my original wording of "fix" :), but either way both should be pretty clear :)

> maybe we should also add a sentence what needs to change for the dataframe to become valid

Thanks for adding this!

PR #155, opened 2024-05-31
`ptable_scatters` allow 3rd data dimension for colormap

### LINKED ISSUE #141: `ptable_scatters` allow 3rd data dimension for colormap
### `ptable_scatters` allow 3rd data dimension for colormap

Remind myself in case it gets lost.

Need to re-enable: https://github.com/janosh/pymatviz/blob/39fd71e5e105574ed1e8d1e3174f9ae4257f95be/tests/test_ptable.py#L528-L540

Also need to fix plotting of `examples/diatomics/homo-nuclear-mace-medium.svg`.

https://github.com/janosh/pymatviz/blob/6ecf68a5f449d8c6b93a16c553d6233433612948/pymatviz/ptable.py#L1586

DanielYang59:
## Summary

- `ptable_scatters` allow 3rd data dimension for colormap, to close #141.
- Fix plotting of `examples/diatomics/homo-nuclear-mace-medium.svg`
- [x] Clarify `PTableProjector` property `elem_colors`: https://github.com/janosh/pymatviz/blob/aaa7aa1bf3508021aae1fd6a5332cda4b71b9669/pymatviz/ptable.py#L300-L304
- Fix a minor issue where the default `text_color` as `element-types` could overwrite default `elem_type_color`: https://github.com/janosh/pymatviz/blob/523101bc6af2e0b9168a1775cd9c95dd09c3ec4b/pymatviz/ptable.py#L405
  ```python
  text_color="element-types" if color_elem_strategy in {"both", "symbol"} else "black"
  ```


pymatviz/ptable.py
```diff
@@ -397,16 +397,20 @@ def add_elem_symbols(
 
             # Get axis index by element symbol
             symbol: str = element.symbol
+            if symbol not in self.data.index:
```
DanielYang59, line 400: Wondering why this part was even working before? As if we don't check if an element is present, the symbol of all elements would be plotted as in 97d453708e53ec377b78a921aa42f5ffdb7870ed

DanielYang59:
@janosh Please review this.

One thing left though, we might need to set a good default for the `elem_colors` (color for each element)? Do you have any advice on this? This sounds a bit tricky as there are too many elements.

https://github.com/janosh/pymatviz/blob/523101bc6af2e0b9168a1775cd9c95dd09c3ec4b/pymatviz/ptable.py#L295-L304

pymatviz/ptable.py
```diff
@@ -1619,28 +1640,28 @@ def ptable_scatters(
...
-        colormap=None,
+        colormap=colormap,
         plot_kwargs=plot_kwargs,  # type: ignore[arg-type]
         hide_f_block=hide_f_block,
         elem_type_colors=elem_type_colors,
     )
 
     # Call child plotter: Scatter
-    plotter.add_child_plots(
+    child_kwargs = {"cmap": projector.cmap, "norm": projector.norm}
```
janosh, line 1654: is the `child_kwargs` keyword of `ptable_scatters` left by accident? currently it's just being overwritten so whatever the user passes makes no difference

assets/ptable-scatters.svg
janosh, line 1: the colors look great! let's make the data slightly less random to make the plot look more like something you would see in the real world. a parity plot with data along the diagonal but adding some Gaussian noise would be good.

janosh (changes requested):
thanks @DanielYang59, great work as usual! 🚀  just 2 minor comments

pymatviz/ptable.py
DanielYang59: Sorry! I must have made a mistake here, thanks for catching this and I would fix this now :)

pymatviz/ptable.py
```diff
@@ -1425,9 +1425,7 @@ def ptable_hists(
     plot_kwargs: dict[str, Any] | None = None,
     # Axis-scope
     ax_kwargs: dict[str, Any] | None = None,
-    child_kwargs: dict[str, Any]
-    | Callable[[Sequence[float]], dict[str, Any]]
```
DanielYang59, line 1424: I'm a bit unsure about `child_kwargs` as `Callable` here

assets/ptable-scatters.svg
DanielYang59: Done! What about this 5efe0495eb934038ee5445707f3973c384c12dbd? Currently it's not shown in the main readme, perhaps add it there?

janosh: https://github.com/janosh/pymatviz/commit/5efe0495eb934038ee5445707f3973c384c12dbd looks great! thank you.

let's keep both in the readme. rather than remove heatmap-3-split i would prefer adding a second `ptable_scatters` to show some of the non-default plot options

pymatviz/ptable.py
janosh: i was confused by the `Callable` type in there as well. i think we can remove it. what was the intended use case anyway? allow people to customize `child_kwargs` based on the data for each element tile?

janosh:
> One thing left though, we might need to set a good default for the elem_colors (color for each element)? 

forgot to reply to this question earlier. how about we copy the VESTA and JMOL color palettes [over from `elementari`](https://github.com/janosh/elementari/blob/85a044cdec93c4eedf3f39d340de41ff90053862/src/lib/colors.ts#L20-L242)? we could then define a `Literal` or `Enum` with special values

```py
elem_colors: Literal["vesta", "jmol"]
```

assets/ptable-scatters.svg
DanielYang59: > [5efe049](https://github.com/janosh/pymatviz/commit/5efe0495eb934038ee5445707f3973c384c12dbd) looks great! thank you.

Very glad to hear that :)

> let's keep both in the readme. rather than remove heatmap-3-split i would prefer adding a second `ptable_scatters` to show some of the non-default plot options

Do you have some advice?

pymatviz/ptable.py
DanielYang59: > i was confused by the `Callable` type in there as well. i think we can remove it. what was the intended use case anyway? allow people to customize `child_kwargs` based on the data for each element tile?

Not sure, I think this is some legacy from previous implementation of `ptable_hists`? Let's remove it for now, maybe add it when we have a solid use case in the future

pymatviz/_colors.py
```diff
@@ -0,0 +1,256 @@
...
+    "Halogen": "teal",
+    "Metal": "lightblue",
+    "Alkaline Metal": "magenta",
+    "Transactinide": "olive",
+}
+
+# The following element-based colors are taken from elementari:
+# https://github.com/janosh/elementari
+ELEM_COLORS_JMOL: dict[str, Color] = {
+    "H": (255, 255, 255),
```
DanielYang59, line 35: Need double check, `matplotlib` color in 0-1 range or 0-255?

janosh: they're 0-1:

```py
from matplotlib.colors import rgb2hex
from matplotlib.typing import ColorType

ColorType
rgb2hex((1,1,1))
```

```py
rgb2hex((255,1,1))
>>> ValueError: RGBA values should be within 0-1 range
```

janosh: i think the `colors` should really be public. i can see use cases where people want to import the JMOL or VESTA element color schemes

janosh (approved):
thanks a lot @DanielYang59, this looks great!

DanielYang59:
Thanks for the final beautiful touch :)

pymatviz/ptable.py
```diff
@@ -299,21 +299,21 @@ def elem_colors(self) -> dict[str, ColorType]:
     @elem_colors.setter
     def elem_colors(
         self,
-        elem_colors: Literal["vesta", "jmol"] | dict[str, ColorType] = "vesta",
+        elem_colors: ElemColors | dict[str, ColorType] | None = ElemColors.vesta,
```
DanielYang59, line 302: Wondering why do we need a `None` as default here? which would resolve to `vesta` anyway?

janosh: i imagine someone who wants to set the colors back to default and doesn't know what the default is might just try

```py
projector.elem_colors = None
```

and see what happens. maybe that's just me in which i'm open to changing this

DanielYang59: > i imagine someone who wants to set the colors back to default and doesn't know what the default is might just try
> 
> ```python
> projector.elem_colors = None
> ```
> 
> and see what happens.

I think if a user don't read the docstring/manual at all, they may not image setting `None` would set the colors to default in the first place (they might expect the default to be a `dict` mapping). On the other hand, if they read the docstring, they would know how to set the colors, so I think `None` adds to ambiguity here :)

janosh: fair, feel free to remove the `None`

DanielYang59: Okay!

PR #157, opened 2024-06-07
[Breaking] Refactor ptable heatmap plotter

### LINKED ISSUE #142: [Enhancement] Refactor`ptable_hists` and `ptable_heatmap` with new `PTableProjector`

DanielYang59:
## Summary

- Refactor ptable heatmap plotter, to close #142.
- Add a new `PTableData` class to handle data preprocessing in a more systematic way for ptable plotters
- Default `on_empty` value is not `show` instead of `hide`, rationale: user might expect a "complete" looking ptable even when data for element is missing

### Update example ipynb
- [x] examples/matbench_dielectric_eda.ipynb
- [x] examples/mprester_ptable.ipynb
- [x] dataset_exploration

### Things for a follow up PR
- [ ] tile size setter is not working properly
- [ ] f_block offset (95cfc76ee3bf84f328fdb24a2862dc4d0d0a67d9 doesn't work)
- [ ] Simplify internal functions for `check_and_replace_missing`, `check_and_replace_infinity` and `log_scale`


janosh:
planning a v0.9.0 release in the coming days (minor version bump due to the breaking changes in #161). anything i can help with here to get this merged before then?

DanielYang59:
> planning a v0.9.0 release in the coming days (minor version bump due to the breaking changes in #161). anything i can help with here to get this merged before then?

Thanks for asking :). Sorry I haven't got much time to work on this now (at home with families). Perhaps we could issue a new release first and I would catch up later. I certainly need some time to polish this refactor ;)

BTW, the parity -> scatter rename is something I want to ask for a while :) I thought you have a reason to name it as `parity` that I'm not aware of.

janosh:
> Sorry I haven't got much time to work on this now (at home with families).

no need to apologize, take your time! 👍 tell your family i said hi 😄

pymatviz/ptable.py
```diff
@@ -703,34 +701,39 @@ def histogram(
...
     zero_symbol: str | float = "-",
-    text_style: dict[str, Any] | None = None,
-    label_font_size: int = 16,
-    value_font_size: int = 12,
-    tile_size: float | tuple[float, float] = 0.9,
+    show_values: bool = True,
+    log: bool | Normalize = False,
+    # Figure-scope
     f_block_voffset: float = 0.5,
     hide_f_block: bool | None = None,
```
DanielYang59, line 721: I kind of want to change `hide_f_block` to something like `bool | Literal["AUTO"] = "AUTO"`, and detect if data if present for f-block by default. I think this is more self-descriptive than `None`?

pymatviz/ptable.py
```diff
@@ -1001,6 +797,7 @@ def ptable_heatmap_splits(
     colormap: str | Colormap = "viridis",
     on_empty: Literal["hide", "show"] = "hide",
     hide_f_block: bool | None = None,
+    # TODO: I think we agreed to remove the "Callable" type of "plot_kwargs"?
     plot_kwargs: dict[str, Any]
     | Callable[[Sequence[float]], dict[str, Any]]
```
DanielYang59, line 802: I think we decided to remove `Callable` from here? Need to double check.

pymatviz/ptable.py
```diff
@@ -538,6 +536,43 @@ class ChildPlotters:
     (line/scatter/histogram) for individual element tiles.
     """
 
+    @staticmethod
+    def heatmap(
```
DanielYang59, line 540: Had to add another `heatmap` child plotter instead of reusing the `rectangle` plotter because `set_clip_path` doesn't seem to show the boarder for each tile (not sure if this is solvable). 

Another reason being I kind of want to integrate some of the processing steps for "percentage/ratio/..." inside the child plotter.

Let's see how everything goes and decide later.

pymatviz/ptable.py
janosh: yes, that's more descriptive 👍

pymatviz/ptable.py
janosh: > because set_clip_path doesn't seem to show the border for each tile (not sure if this is solvable).

not sure, maybe something to ask an LLM 😄

pymatviz/ptable.py
janosh: yes, i also remember discussing that

DanielYang59: Thanks for confirming. I would remove this :)

pymatviz/ptable.py
```diff
@@ -703,293 +747,154 @@ def histogram(
...
-    ax: plt.Axes | None = None,
-    count_mode: ElemCountMode = ElemCountMode.composition,
+    # TODO: to be removed
+    count_mode: ElemCountMode = ElemCountMode.composition,  # separate data preprocess stage
+    # Heatmap specific
+    heat_mode: Literal["value", "fraction", "percent"] | None = "value",
+    # infty_color: str = "lightskyblue",
+    # na_color: str = "white",
+    # zero_color: str = "#eff",  # light gray  # TODO:covered by colormap?
+    # zero_symbol: str | float = "-",
```
DanielYang59, line 759: Not sure if we need to treat zeros specially here (especially for the color, which should be covered by the colormap?)?

pymatviz/ptable.py
```diff
@@ -703,293 +752,210 @@ def histogram(
...
-    cbar_range: tuple[float | None, float | None] | None = None,
+    # cbar_range: tuple[float | None, float | None] | None = None,
     cbar_coords: tuple[float, float, float, float] = (0.18, 0.8, 0.42, 0.05),
+    cbar_title_kwargs: dict[str, Any] | None = None,
     cbar_kwargs: dict[str, Any] | None = None,
-    colorscale: str = "viridis",
-    show_scale: bool = True,
-    show_values: bool = True,
-    infty_color: str = "lightskyblue",
-    na_color: str = "white",
```
DanielYang59, line 719: Need some discussion on NaN and infinity colors here, as the new implementation would replace them. Do we want to treat them specially or keep consistent with the new plotter?

Meanwhile I think passing NaN or inf should be the edge case for heatmap plotter?

janosh: for `ptable_heatmap_ratio` the objective is definitely to show 0 and inf as anomalous values. if you're looking at the ratio of two element distributions 0 means an element is not in the numerator, whereas inf means it's not in the denominator which is worth highlighting with separate colors. if you're not doing a ratio plot, 0 is not important to highlight and inf should very rarely happen but still worth having a special color for inf imo to cover edge cases and point out to the users there's something anomalous in the heatmap values

DanielYang59: Thanks for the input. Yes it's always good to have the flexibility to assign a special color to a special value (inf/NaN/zero in this case). 

I need to have a good think about how to actually achieve this :) Perhaps we need another preprocessor for heatmap plotter (which should be pretty easy, because it's just a `{element: single value}` pair). I would take care of this.

Meanwhile, while I'm working on this, I noticed my original plan (migrate everything at the same time) doesn't seem to work because there are simply too many functionalities to rewrite. I would instead refactor with the base function, and add one functionality at a time (see the top of the PR) :)

And here is what we got right now:
ptable_test

pymatviz/ptable.py
DanielYang59: I would fold this for now as this would be included in the special value handler plan (i.e. select custom color for a particular value, NaN/inf/zero and more).

pymatviz/ptable.py
```diff
@@ -703,293 +751,208 @@ def histogram(
...
-        colorscale (str, optional): Matplotlib colormap name to use. Defaults to
-            "viridis". See https://matplotlib.org/stable/users/explain/colors/colormaps
-            for available options.
-        show_scale (bool, optional): Whether to show the color bar. Defaults to True.
-        show_values (bool, optional): Whether to show the heatmap values in each tile.
-            Defaults to True.
-        infty_color: Color to use for elements with value infinity. Defaults to
-            "lightskyblue".
-        na_color: Color to use for elements with value infinity. Defaults to "white".
-        heat_mode ("value" | "fraction" | "percent" | None): Whether to display heat
```
DanielYang59, line 765: @janosh Can I have a quick explaination on this argument please?

https://github.com/janosh/pymatviz/blob/d222982c423a78e69914f3ae67942bc44ecf10ba/pymatviz/ptable.py#L765-L769

What is the difference between "fraction" mode and "percent"?

Any what does `None` do? Does `None` hide the values altogether, but we already have `show_values` for this?

janosh: `percent` and `fraction` only differ by a factor of 100 and a suffix of `%` (IIRC). so `fraction` would render as `0.025` and `percent` as `2.5%`. and yes, `None` should hide the values altogether. it's possible that that's redundant with `show_values` now. i may have forgotten about the `heat_mode=None` option, when i added `show_values` 😅

pymatviz/ptable.py
janosh: looks good, thanks for sharing! are there any visual differences compared to before? maybe post the same plot generated from `main` for comparison?

janosh: btw, thanks for working on this! good timing as i was planning to make some changes to `ptable_heatmap_plotly` this weekend which might have caused merge conflicts here (though probably not) so i'll hold off on that just in case

pymatviz/ptable.py
DanielYang59: BTW, what do you think of the idea to rename `heat_mode` (which is a bit ambiguous to me personally) to `values_mode` (for "values display mode")? Now that we are grouping arguments.
```python
# Values
show_values: bool = True,
heat_mode: Literal["value", "fraction", "percent"] = "value",  # values_mode maybe?
values_pos: tuple[float, float] | None = None,
values_color: str = "black",  
values_kwargs: dict[str, Any] | None = None,
```

pymatviz/ptable.py
DanielYang59: > looks good, thanks for sharing! 

Thanks!

> are there any visual differences compared to before? maybe post the same plot generated from `main` for comparison?

I haven't done this comparison yet. Would do tomorrow :)

pymatviz/ptable.py
janosh: i'm definitely open to a change. we can make it backwards compatible with a deprecation warning by still looking for `heat_mode` in `kwargs` and renaming it to `values_display_mode` if found.

though come to think of it, maybe `show_value_mode` is similarly clear and a little shorter than `values_display_mode`. `show_value_mode` could then entirely replace `show_values`. the previous `show_values=True|False` would become `show_value_mode='value'|'off'`. so in total, we'd have `show_value_mode: Literal["value", "fraction", "percent", "off"]`

DanielYang59: Yes! Let's combine these two args into a single one :)

pymatviz/ptable.py
DanielYang59: Here is the heatmap from the plotter in main:

There are a few minor differences:
- The font looks like bold by default, and in "white" color of course.
- The color of the values would switch between black and white automatically depending on the background color, that's a good feature to have. I would migrate this.
- The default decimal places are a bit different, but still should be an easy fix.

ptable_test

pymatviz/ptable.py
```diff
@@ -703,293 +766,292 @@ def histogram(
...
+                text_color (str): Value text color.
+                kwargs (dict): Additional keyword arguments to pass to the `ax.text`.
+            """
+            # Update symbol kwargs
+            kwargs = kwargs or {}
+            kwargs.setdefault("fontsize", 12)
+
+            # Generate values format depending on the display mode
+            if text_fmt == "AUTO":
+                text_fmt = ".1%" if values_show_mode == "percent" else ".3g"
```
DanielYang59, line 958: Here we have some inconsistency, especially with the `normalize_data` util, as `text_fmt` as `".1%"` would multiply data by 100. Perhaps instead of taking the format f-string, we ask for the total number of digits?

janosh: i really like the f-string float format syntax because its so versatile. has way more features than just precision like left/right alignment, zero padding, thousand separator, percentages, etc. so would always piggy back on that where possible over restricting user control to only precision

DanielYang59: Okay sure no problem :) Let's use f-string. That's always a hard choice, where more flexible means more possibilities to trigger expected behavior :)

pymatviz/ptable.py
```diff
@@ -463,8 +477,14 @@ def add_colorbar(
         # Add colorbar
         cbar_ax = self.fig.add_axes(coords)
 
+        # Set colorbar range
+        self._norm = Normalize(
+            vmin=cbar_range[0] or self.data.attrs["vmin"],
+            vmax=cbar_range[1] or self.data.attrs["vmax"],
```
DanielYang59, line 483: Need confirmation on this to ensure re-scale the normalizer would not have any side effect on value-color mapping.

pymatviz/ptable.py
```diff
@@ -700,291 +763,305 @@ def histogram(
...
-            label = r"$\infty$"
-        elif pd.isna(tile_value):
-            color = na_color  # neither numerator nor denominator
-            label = r"$0\,/\,0$"
-        elif tile_value == 0:
-            color = zero_color
-            label = str(zero_symbol)
-        else:
-            color = color_map(norm(tile_value))
+    projector.add_child_plots(
```
DanielYang59, line 1005: I'm thinking about replacing this calling child plots architecture right now, because it makes adding custom child plotters really hard and anti-intuitive. 

Instead of writing a child plotter, pass it to the plot caller, we might just inherit from the `PTableProjector` and overwrite the `add_child_plots` method.

pymatviz/ptable.py
```diff
@@ -700,291 +765,352 @@ def histogram(
...
     *,
-    log: bool | Normalize = False,
-    ax: plt.Axes | None = None,
-    count_mode: ElemCountMode = ElemCountMode.composition,
-    cbar_title: str = "Element Count",
-    cbar_range: tuple[float | None, float | None] | None = None,
+    # Heatmap specific
+    colormap: str = "viridis",
+    log: bool = False,
+    sci_notation: bool = False,
```
DanielYang59, line 773: I think we might just make our life easier to add this `sci_notation` arg to manually toggle scientific notation mode, instead of trying to guess whether we need scientific notation mode or not (because I cannot think about a deterministic and algorithmic way to decide this).

pymatviz/ptable_test_percentage.png
DanielYang59, line 1: I committed these three examples for you to keep track of where we are right now :) Would certainly remove them before merge.

pymatviz/ptable.py
```diff
@@ -699,292 +730,438 @@ def histogram(
...
-    colorscale: str = "viridis",
-    show_scale: bool = True,
-    show_values: bool = True,
-    infty_color: str = "lightskyblue",
-    na_color: str = "white",
-    heat_mode: Literal["value", "fraction", "percent"] | None = "value",
-    fmt: str | Callable[..., str] | None = None,
-    cbar_fmt: str | Callable[..., str] | None = None,
-    text_color: str | tuple[str, str] = "auto",
-    exclude_elements: Sequence[str] = (),
```
DanielYang59, line 721: This arg `exclude_element` is now removed, the rationale being: I think it's something that should be controlled by the `data` itself. I.e. if an element is not present in the data, then it would not be included.

pymatviz/ptable.py
```diff
@@ -699,292 +730,438 @@ def histogram(
         ax.spines[["right", "top", "left"]].set_visible(False)
 
 
-def ptable_heatmap(
-    values: ElemValues,
-    *,
-    log: bool | Normalize = False,
-    ax: plt.Axes | None = None,
-    count_mode: ElemCountMode = ElemCountMode.composition,
```
DanielYang59, line 707: I'm a bit unsure about `count_mode` at this moment. I think for our current implementation, the `data` is expected to be an element to value mapping.

It looks like this (`count_elements`) is a preprocessing function, perhaps we should separate the preprocessing stage from the plotter?

pymatviz/ptable.py
```diff
@@ -699,292 +730,438 @@ def histogram(
...
-    show_scale: bool = True,
-    show_values: bool = True,
-    infty_color: str = "lightskyblue",
-    na_color: str = "white",
-    heat_mode: Literal["value", "fraction", "percent"] | None = "value",
-    fmt: str | Callable[..., str] | None = None,
-    cbar_fmt: str | Callable[..., str] | None = None,
-    text_color: str | tuple[str, str] = "auto",
-    exclude_elements: Sequence[str] = (),
-    zero_color: str = "#eff",  # light gray
```
DanielYang59, line 722: I think we might just migrate this `zero_color` and `zero_symbol` to `ptable_heatmap_ratio` as I think zero/inf should be a common case for ratio plotters (perhaps not the ptable heatmap plotter). Perhaps also add a `zero_tolerance` arg?

 If they user really want to overwrite the color for zero, they could use `overwrite_colors` to overwrite that.

tests/ptable/test_ptable_matplotlib.py
```diff
@@ -154,155 +154,88 @@ def steel_elem_counts(steel_formulas: pd.Series[Composition]) -> pd.Series[int]:
...
-    check_text_color_consistency(ax)
-
+    def test_values_show_mode(
+        self, values_show_mode: Literal["percent", "fraction", "value", "off"]
+    ) -> None:
+        fig = ptable_heatmap(df_ptable.atomic_mass, values_show_mode=values_show_mode)
+        assert isinstance(fig, plt.Figure)
+        assert len(fig.axes) == 181
 
-def check_text_color_consistency(ax: plt.Axes) -> None:
```
DanielYang59, line 270: I would remove this for now because with current implementation symbols and values would take color from the same mapping, so would expect them to be always consistent. 

Mark this because it's always better to have redundancy in tests, but I currently cannot think of a proper way to get the colors of values/symbols.

DanielYang59:
@janosh  Finally cleared my TODO list, and I didn't find any more issues to fix for now (except for the tile size and position one which is planned for a future PR). Please review this at your convenience, thanks!

I do notice the one unit test failure (passed locally), but I'm not sure if it's related to this PR. I would have a closer look later.
```
=================================== FAILURES ===================================
______________________ test_count_elements_by_atomic_nums ______________________
    def test_count_elements_by_atomic_nums() -> None:
        series_in = pd.Series(1, index=range(1, 119))
>       el_cts = count_elements(series_in)
tests/test_process_data.py:29: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
pymatviz/process_data.py:107: in count_elements
    df_ptable.reset_index().set_index("atomic_number").symbol
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
self =               Key.element  row  column  ...         jmol_color  covalent_radius  tmp
atomic_number                    ... NaN    1
118                    Og    7      18  ...                NaN              NaN    1
[118 rows x 32 columns]
name = 'symbol'
    @final
    def __getattr__(self, name: str):
        """
        After regular attribute access, try looking up the name
        This allows simpler access to columns for interactive use.
        """
        # Note: obj.x will always call obj.__getattribute__('x') prior to
        # calling obj.__getattr__('x').
        if (
            name not in self._internal_names_set
            and name not in self._metadata
            and name not in self._accessors
            and self._info_axis._can_hold_identifiers_and_holds_name(name)
        ):
            return self[name]
>       return object.__getattribute__(self, name)
E       AttributeError: 'DataFrame' object has no attribute 'symbol'
/opt/hostedtoolcache/Python/3.9.19/x64/lib/python3.9/site-packages/pandas/core/generic.py:6299: AttributeError
```

examples/dataset_exploration/matpes/eda.py
```diff
@@ -26,6 +26,7 @@
 
 
 # %%
+# DEBUG: following data doesn't exist
 with gzip.open(f"{module_dir}/20240214-MatPES-178070-r2SCAN.json.gz", "rt") as file:
```
DanielYang59, line 30: Not sure where to download this dataset?

janosh: yes, that dataset is not public yet. i should have added a comment about that. [just did so now](https://github.com/janosh/pymatviz/commit/b6cdca34a3643aa9dc04070bb92470bb20e7ad3e#diff-2490fe956ff43be755c667a04bdfee7e0d1d6a13d8c92a62175d5ab33e94d826R28-R30)

pymatviz/utils.py
```diff
@@ -362,6 +370,59 @@ def si_fmt(
 si_fmt_int = partial(si_fmt, fmt=".0f")
 
 
+def get_cbar_label_formatter(
```
janosh, line 373: are you sure we need to make this a function? might make sense to inline this code since it seems to only be used one and the code isn't that complex

examples/dataset_exploration/ricci_carrier_transport/explore_carrier_transport.py
```diff
@@ -23,24 +23,29 @@
...
 
 # %%
 df_carrier = load_dataset("ricci_boltztrap_mp_tabular")
 
-# getting space group symbols and numbers takes about 2 min
+# TODO: Many structure is a float for some reason?
+df_carrier = df_carrier[
+    df_carrier[Key.structure].apply(lambda x: not isinstance(x, float))
+]
+
```
janosh, line 38: actually, these structures are not floats, they are missing which is interpreted as `float("nan")`. a cleaner way to filter them out would be `df_carrier.dropna(subset=Key.structure)`

pymatviz/ptable/ptable_matplotlib.py
```diff
@@ -1,888 +1,407 @@
...
+    cbar_title_kwargs: dict[str, Any] | None = None,
+    cbar_kwargs: dict[str, Any] | None = None,
+    # Deprecated args, don't use
+    colorscale: str | None = None,
+    heat_mode: Literal["value", "fraction", "percent"] | None = None,
+    show_values: bool | None = None,
+    fmt: str | None = None,
+    cbar_fmt: str | None = None,
+    show_scale: bool | None = None,
+) -> plt.Figure:
```
janosh, line 81: i think returning a `plt.Figure` instead of `plt.Axes` is a big breaking change. not something to do lightly. i realize it makes the `matplotlib` `ptable` plotters more consistent, but a lot of the `pymatviz` code i and others like @CompRhys wrote relies on `ptable_heatmap` returning `Axes`. at the least, i think we would need to try to make an automatic migration.

janosh (changes requested):

examples/dataset_exploration/matpes/eda.py
DanielYang59: Thanks for the input!

examples/dataset_exploration/ricci_carrier_transport/explore_carrier_transport.py
DanielYang59: Thanks for the input. Sorry I didn't check the actual value but the type. I would drop NaN then :)

pymatviz/ptable/ptable_matplotlib.py
DanielYang59: Okay! Sure. I would revert this change (as current projector implementation would return the `Figure`), I would have a look and try to make it backwards compatible

pymatviz/utils.py
DanielYang59: I think the reason I made some function that could be done inline is that I really want them to be well covered by unit tests :)

I have no objection to make this inline, but I reaaaaally don't want to lose those unit tests. I would appreciate it if you could give me some advice on this :)

And I'm hoping this could be useful for other plotters which include a colorbar.

pymatviz/ptable/ptable_matplotlib.py
DanielYang59: I just added a `return_type` (default to `axes` to be backwards compatible) in 0f9bdc796f8f90b6c92d344e062a4d2479d75117 to encourage the migration from `axes` to `figure` within one year period, hopefully this could make this transition smoother. 

I have to note that because of the change in `axes` order, the resulted figure could be suboptimal though:
```python

from pymatviz.enums import Key
from pymatviz.io import save_and_compress_svg
from pymatviz.process_data import count_elements
from pymatviz.ptable import ptable_heatmap


df_expt_gap = load_dataset("matbench_expt_gap")


# Elemental Plots
ax = ptable_heatmap(count_elements(df_expt_gap[Key.composition]), log=True)
title = (
    f"Elements in Matbench Experimental Band Gap ({len(df_expt_gap):,} compositions)"
)
ax.set_title(title, y=0.96, fontsize=16, fontweight="bold")
save_and_compress_svg(ax, "ptable-heatmap")
```

Where the title kind of overlap with the body of the figure (changing the `y` doesn't seem to move the title, I guess it's owing to the position of the ax).

ptable-heatmap

janosh: i just tried it. changing `y` in `ax.set_title()` still has an effect but it's much less pronounced than before as the units are relative to the size of the axes (which before was the whole figure and now is just one element tile)

DanielYang59: Yes exactly, it seems the new implementation changed the "scope" of the axes. Sorry it looks like a wrong timing to propose this breaking change (though the `return_type` arg should make it non-breaking, but less pretty).

In current implementation, the `ax` is generated via: https://github.com/janosh/pymatviz/blob/2c4c5da15ea1ffe6cb85f1092b0ab8519c73eb94/pymatviz/ptable/ptable_matplotlib.py#L737-L739

With the new implementation, we now switch to `plt.subplots` to generate the `figure` and `axes`:
```python
self.fig, self.axes = plt.subplots(n_periods, n_groups, **plot_kwargs)
```

Upon a closer look, I was able to "fix" the inconsistency by using a huge relative position:
```python

from pymatviz.enums import Key
from pymatviz.io import save_and_compress_svg
from pymatviz.process_data import count_elements
from pymatviz.ptable import ptable_heatmap


df_expt_gap = load_dataset("matbench_expt_gap")


# Elemental Plots
ax = ptable_heatmap(count_elements(df_expt_gap[Key.composition]), log=True)
title = (
    f"Elements in Matbench Experimental Band Gap ({len(df_expt_gap):,} compositions)"
)
ax.set_title(title, x=0.75, y=2.5, fontsize=18, fontweight="bold")
save_and_compress_svg(ax, "ptable-heatmap")
```
debug

So I would say the new implementation is still "compatible" with previous version (at the price of having to tweak the title position), but I wish to encourage users to migrate to `figure` over the one year period.

Do you have further comment or can you review (and possibly merge) this? Thanks!

tests/test_process_data.py
```diff
@@ -24,6 +24,7 @@ def test_count_elements(count_mode: ElemCountMode, counts: dict[str, float]) ->
     pd.testing.assert_series_equal(series, expected, check_dtype=False)
 
 
+# DEBUG: this test is failing in CI (not locally)
```
DanielYang59, line 27: I totally have no idea why this unit test is failing (in CI only, passed locally):
```python
def test_count_elements_by_atomic_nums() -> None:
    series_in = pd.Series(1, index=range(1, 119))
    el_cts = count_elements(series_in)
    expected = pd.Series(1, index=df_ptable.index, name="count")

    pd.testing.assert_series_equal(expected, el_cts)
```

As `ptable.process_data` is not touched during this PR at all...

tests/ptable/test_ptable_matplotlib.py
```diff
@@ -26,6 +26,8 @@
 
     from pymatgen.core import Composition
 
+df_ptable = df_ptable.copy()  # avoid changing the df in place
```
DanielYang59, line 29: Don't know if there will be a better solution (`pandas` doesn't seem to provide any method to make a dataframe immutable)?

Also this is pretty interesting (my first time to notice one unit test alters the behavior of another), in the case `test_ptable_matplotlib.py` seems to alter `df_table` and therefore `test_process_data.py` failed (just guessing).

janosh (approved):
thanks a lot @DanielYang59! 👍 i appreciate all the work that went into this PR!

apologies it took this long too review. there's something to be said for small PRs since it's hard to find a sufficiently long stretch of free time to review large PRs in one go other than weekends. 😄

DanielYang59:
No need to apologize in any case, and thanks for reviewing and the final top up! I really appreciate your time from your weekend to review this. I'm just hoping not too many things would be broken by this PR :)

Meanwhile wondering why we're falling back from `isclose` to full equality in 34713e8e410df98df7559304a4aa751115882a4b? The metadata is exactly the same at this moment, but I'm not quite sure if we were to introduce any formatting mechanism (e.g. from `1` to `1.0`) in the future which might make full equality comparison unsuitable.

janosh:
> The metadata is exactly the same at this moment, but I'm not quite sure if we were to introduce any formatting mechanism (e.g. from 1 to 1.0) in the future which might make full equality comparison unsuitable.

in this case i don't mind for the tests to be sensitive. if a refactor changes exact equality, i think it's good for the tests to alert us to this even if we go with it and then need to update the tests

DanielYang59:
Sure thing!

PR #158, opened 2024-06-11
Fix `ptable_heatmap_ratio` legend

janosh:
**before**
broken-ptable-heatmap

**after**
fixed-ptable-heatmap


assets/ptable-scatters-parity.svg
DanielYang59, line 1: @janosh I'm afraid the scatter examples are broken for some reason?

janosh: good catch, that must be some bad SVG compression by [vecta.io/nano](https://vecta.io/nano)

PR #162, opened 2024-06-22
Fix `ruff` `NPY002`

janosh:


examples/make_assets/ptable.py
```diff
@@ -86,9 +87,7 @@
 
 # %% Histograms laid out in as a periodic table
 # Generate random parity data with y \approx x with some noise
-data_dict = {
-    elem.symbol: np.random.randn(100) + np.random.randn(100) for elem in Element
-}
+data_dict = {elem.symbol: np_rng.randn(100) + np_rng.randn(100) for elem in Element}
```
DanielYang59, line 90: @janosh It's so interesting, as I started working on fixing `NPY002` violations in `pymatgen` https://github.com/materialsproject/pymatgen/pull/3892 without looking at this PR almost at the same time (really!).

But this legacy to new Generator replacement is not just replacing `np.random` with `rng` (the grammar is not exactly the same, please refer to https://numpy.org/doc/stable/reference/random/generator.html), this should be `np_rng.standard_normal(100)` instead.

examples/make_assets/sankey.py
```diff
@@ -8,15 +8,16 @@
 
 # %% Sankey diagram of random integers
 cols = ["col_a", "col_b"]
-df_rand_ints = pd.DataFrame(np.random.randint(1, 6, size=(100, 2)), columns=cols)
+np_rng = np.random.default_rng(seed=0)
+df_rand_ints = pd.DataFrame(np_rng.randint(1, 6, size=(100, 2)), columns=cols)
```
DanielYang59, line 12: I haven't got time to look through everything yet (perhaps would do tomorrow if you want), this should be `np_rng.integers()`.

examples/make_assets/uncertainty.py
```diff
@@ -27,7 +28,7 @@
 ax = error_decay_with_uncert(y_true, y_pred, y_std)
 save_and_compress_svg(ax, "error-decay-with-uncert")
 
-eps = 0.2 * np.random.randn(*y_std.shape)
+eps = 0.2 * np_rng.randn(*y_std.shape)
```
DanielYang59, line 31: Same here, should be `rng.standard_normal()`

tests/conftest.py
```diff
@@ -26,14 +26,14 @@
 
 
 # random regression data
-np.random.seed(42)
-xs = np.random.rand(100)
-y_pred = xs + 0.1 * np.random.normal(size=100)
-y_true = xs + 0.1 * np.random.normal(size=100)
+np_rng = np.random.default_rng(seed=0)
+xs = np_rng.uniform(0, 1, 100)
```
DanielYang59, line 30: I guess this is the same as `np_rng.random(100)`?

examples/make_assets/ptable.py
janosh: good catch! thanks for letting me know. i should really take the time to make these scripts run in CI to avoid such mistakes...

will submit a fix now

DanielYang59: No problem at all :) Yes we need to include them into CI tests, mind letting me (learn and) do this?

PR #176, opened 2024-07-09
Moving enums may have broken end users pickle's, reduce to str when pickling to be more backwards compatible going forward.

CompRhys:
Overwrites the `__reduce_ex__` method in `LabelEnum`


janosh (approved):
thanks @comprhys!

PR #180, opened 2024-07-17
Fix ptable scatter examples in homepage

DanielYang59:
### Summary

- Fix ptable scatter examples in homepage

I'm sure @janosh fixed this before but for some reason it's now broken again


janosh (approved):
good catch, thanks @DanielYang59!

DanielYang59:
No problem, thanks for reviewing!

.github/workflows/svgo.yml
```diff
@@ -24,7 +24,7 @@ jobs:
 
       - name: Compress SVG assets and commit to current PR
         run: |
-          svgo --multipass assets
+          svgo --multipass --final-newline assets --exclude ptable-scatters-*.svg
```
DanielYang59, line 27: @janosh Wondering why do we exclude ptable scatter here? `svgo` seem to work just fine on my machine.

janosh: `svgo` is the SVG compressor that kep repeatedly mangling those exmaple plots you fixed in this PR. like you wrote above, i fixed them before

DanielYang59: I just looked into the history, you indeed fixed them in eb8fd62195d412bbb995913d31b0e908c9c9378d, and for some reason they got broken again in 3e89e590e3f34333731101e44cd75cb9cfdd1eb3. Meanwhile I didn't notice `svgo` to break these two plots on my machine? Perhaps you should try to update `svgo` on your side?

I'm just wondering why `svgo` only break scatter plots, not others 🤣

janosh: the commit that broke the assets was https://github.com/janosh/pymatviz/pull/165/commits/d218a468a4baee2ca3e29a6521df265da204d6f7 which was created by the `svgo.yml` CI workflow, not on my local machine. it always uses the latest `svgo` (running on Linux which shouldn't matter much since SVGO uses nodejs and so should be largely platform independent).

janosh: > I'm just wondering why svgo only break scatter plots, not others 🤣

that i don't know. i suspect it's a bug in `svgo`

DanielYang59: Beautiful. Thanks for the input. I might keep this in to TODO list and perhaps report this to the `svgo` team.

janosh: that would be great! could also be a bug in `matplotlib` btw. perhaps `matplotlib` is not following the SVG spec in some way which makes it harder for `svgo` to losslessly compress. seems less likely though given the SVG renders fine before compression

DanielYang59: Okay! But I would need to figure out how to recreate this in the first place :)

PR #187, opened 2024-07-29
Fix `svgo` workflow for ptable scatter plots

### LINKED ISSUE #184: `svgo` might break ptable scatter plots
As discussed in https://github.com/janosh/pymatviz/pull/180#discussion_r1682934943, `svgo` workflow might break ptable scatter plots. https://github.com/janosh/pymatviz/blob/19fe1c9cc1e5de15bcc43fcaf1c090e0286fdfd0/.github/workflows/svgo.yml#L25-L36

**Need to recreate this issue in the first place.**

DanielYang59:
### Summary

- Allow `svgo` compress workflow for ptable scatter plots, to close #184
- Display `svgo` version for easier tracking of future issue (hopefully never)

The original commit when the issue was first noticed: https://github.com/janosh/pymatviz/pull/165


DanielYang59:
I cannot recreate this issue for some reason, tried to enable `svgo` workflow for scatters in this PR, but didn't see the scatter plots being corrupted.

janosh:
maybe because it's a newer version of `svgo` and they fixed the issue in the meantime? if that's the case, we can remove the `--exlude` flag

DanielYang59:
> maybe because it's a newer version of `svgo` and they fixed the issue in the meantime? 

No idea. And the workflow doesn't show `svgo` version so I cannot recreate that (by using the same version as that broken plots). Therefore added a debug command to show `svgo` version when installed. Hopefully it would help us catch that `svgo` version should this happen again, or if don't it's better and we could remove that tag.

DanielYang59:
@janosh  Perhaps we could merge this now until (hopefully never) this issue pop up again?

.github/workflows/svgo.yml
```diff
@@ -19,18 +19,22 @@ jobs:
...
+          svgo --multipass --final-newline assets
           if git diff --quiet assets; then
             echo "No changes to commit"
             exit 0
           fi
           git config user.name "Janosh Riebesell"
           git config user.email janosh.riebesell@gmail.com
           git add assets
-          git commit -m "compress new SVG assets"
+          git commit -m "[workflow] Compress new SVG assets"
```
DanielYang59, line 39: Added this prefix to distinguish between a commit added by automatic workflow (I thought the original commit was done by you yourself)

janosh: good idea

DanielYang59:
Thanks!

PR #189, opened 2024-07-31
Breaking: drop `plot_` prefix from multiple functions

janosh:
rationale: more consistent since not all functions had the prefix to begin with, should have been all or none. in the future, people might import `pymatviz as pmv` and call e.g. `pmv.phonon_bands()` in which case the `plot_` feels redundant. so to reach consistency, seems better to remove `plot_` prefix than to add it


janosh:
3fcd971 is a minor break as the functions were more conveniently importable from `pymatviz` (via `__init__.py` re-export) and i don't think they had much of a user base yet anyway

PR #192, opened 2024-08-07
`bin_df_cols` leave input df unchanged

janosh:


pymatviz/process_data.py
```diff
@@ -21,7 +21,7 @@ def count_elements(
     values: ElemValues,
     count_mode: ElemCountMode = ElemCountMode.composition,
     exclude_elements: Sequence[str] = (),
-    fill_value: float | None = 0,
+    fill_value: float | None = None,
```
DanielYang59, line 24: This seem to cause the behaviour change in #215, is this an accidental change? I think the original default value of `0` is correct here? i.e. using `0` for element this is missing?

janosh: thanks for looking into this! 🙏 i think returning `None` for missing elements in `count_elements` is more correct/semantic than 0. so ideally, `ptable_heatmap_ratio` should be updated to work with the new behavior

DanielYang59: > i think returning None for missing elements in count_elements is more correct/semantic than 0

I guess both choices hold merit from certain standpoint, i.e. `None` is more semantic to suggest that particular element is missing, `0` is also valid as its count is indeed zero. 

For the particular function, however, I slight prefer using zero because it seems to be designed to "count element occurrence": https://github.com/janosh/pymatviz/blob/82c087b2edaca6c31ba2a97dbb3c9a9114b9425c/pymatviz/process_data.py#L26-L28

So its occurrence would be zero if that particular element doesn't appear in the data.

---

> ideally, ptable_heatmap_ratio should be updated to work with the new behavior

I'm afraid this would have broader impact as there're quite a few usage that need to be checked:
image

For example:
```python
from pymatviz import count_elements
from pymatviz.enums import ElemCountMode, Key


dataset = load_dataset("matbench_expt_gap")[Key.composition]
counted_dataset = count_elements(dataset, ElemCountMode.composition, fill_value=0)
print(counted_dataset)
```

With fill value zero:
```
symbol
H      235.00
He       0.00
Li     967.66
Be     121.00
B     1168.00
       ...   
Fl       0.00
Mc       0.00
Lv       0.00
Ts       0.00
Og       0.00
```

With `None`:
```
symbol
H      235.00
He        NaN
Li     967.66
Be     121.00
B     1168.00
       ...   
Fl        NaN
Mc        NaN
Lv        NaN
Ts        NaN
Og        NaN
```

Using `None` seem to require an additional process step .

PR #194, opened 2024-08-17
Self-import refactor

janosh:
import `pymatviz` as namespace


janosh:
@DanielYang59 curious what you think of refactoring imports in this way, i.e. just having `import pymatviz as pmv` at the top and accessing plot functions and utilities as attributes?

DanielYang59:
Hi @janosh thanks for pinging me. I'm in favor of this way of import: 
1. It kinds of following the similar pattern as `matplotlib` so it would be more intuitive for users to get their hands on hopefully, also performance IMO would not be a key concern for plotting packages (if there is any performance impact).
2. I personally prefer to de-clutter the namespace (especially when some function is used very few times), e.g. in previous code refactoring of `pymatgen` mostly, I would prefer `import math; math.sin/math.cos...` over `from math import sin/cos...`  where the latter brings a lot of functions into the local namespace.
3. I believe we still could keep both ways of import and let user choose whatever they prefer?

DanielYang59:
Thanks for the effort. That's a lot of work!

janosh:
not that much actually. 😄  writing lots of new tests in #193 was much more work

DanielYang59:
I appreciate your effort in both cases :)

PR #196, opened 2024-09-01
ruff fixes

DanielYang59:
## ruff fixes
- Fix T201, replace `print` with others if possible
- Globally ignore SLF001, access private class member 

### TODOs:
- [x] Fix PT017 in `test_io`
- [x] Fix NPY201 in `uncertainty` (this might be bug, if user use NumPy 2)
- [x] Fix PT011, exception capture without `match`
- [x]  "PT006",   # pytest-parametrize-names-wrong-type
- [x] "COM812",  # trailing comma missing
- [x] remove `no_type_check`
- [x] [**need confirm**] CI test reduce items shown in `slowest durations` to 20


pymatviz/phonons.py
```diff
@@ -121,11 +125,7 @@ def get_band_xaxis_ticks(
     return ticks_x_pos, tick_labels
 
 
-YMin = float | Literal["y_min"]
-YMax = float | Literal["y_max"]
-
-
-@no_type_check
+@no_type_check  # TODO: fix this
```
DanielYang59, line 128: fix this in this or a follow up PR

pymatviz/uncertainty.py
```diff
@@ -84,7 +92,7 @@ def qq_gaussian(
         ax.fill_between(
             exp_proportions, y1=obs_proportions, y2=exp_proportions, alpha=0.2
         )
-        miscal_area = np.trapz(  # noqa: NPY201
```
DanielYang59, line 87: Looks like this line is not covered by unit test, this should fail in NumPy 2.x because `np.trapz` is removed.

pyproject.toml
```diff
@@ -97,18 +97,18 @@ ignore = [
...
-    "FIX002",
-    "ISC001",
+    "FIX002",  # line-contains-todo
     "PLR0911", # too-many-return-statements
     "PLR0912", # too-many-branches
     "PLR0913", # function-with-too-many-arguments
     "PLR0915", # too-many-statements
     "PLR2004", # magic-number-comparison
-    "PT006",   # pytest-parametrize-names-wrong-type
+    "PT006",   # pytest-parametrize-names-wrong-type  TODO: fix this
```
DanielYang59, line 106: Tag myself, this should be fixable

tests/ptable/test_ptable_matplotlib.py
```diff
@@ -155,12 +155,12 @@ def test_return_type(self) -> None:
 def test_ptable_heatmap_splits(hide_f_block: bool) -> None:
     """Test pmv.ptable_heatmap_splits with arbitrary data length."""
     data_dict: dict[str, Any] = {
-        elem.symbol: [
-            random.randint(0, 10)  # random value for each split
```
DanielYang59, line 159: Not sure why this triggered [S311 (Checks for uses of cryptographically weak pseudo-random number generators.)](https://docs.astral.sh/ruff/rules/suspicious-non-cryptographic-random-usage/), might need a closer look and report this to ruff if it's a false positive

examples/make_assets/ptable/ptable_matplotlib.py
```diff
@@ -140,8 +140,7 @@
 for n_splits in (2, 3, 4):
     data_dict = {
         elem.symbol: [
-            random.randint(10 * n_splits, 20 * (n_splits + 1))
-            for _ in range(n_splits)  # noqa: S311
```
DanielYang59: Looks like a small issue with `ruff-format` here, causing an interlock:
1. It would first unwrap these two lines, but lose the `noqa` mark.
2. After manually adding back the `noqa` mark to the single line, it would get wrapped again.

pymatviz/phonons.py
```diff
@@ -135,9 +134,15 @@ def _shaded_range(
         return fig
 
     shade_defaults = dict(layer="below", row="all", col="all")
-    y_lim = dict(zip(("y_min", "y_max"), fig.layout.yaxis.range, strict=True))
+    y_lim: dict[float | Literal["y_min", "y_max"], Any] = dict(
+        zip(("y_min", "y_max"), fig.layout.yaxis.range, strict=True),
+    )
 
+    # DEBUG: why (0, "y_min")
     shaded_ys = shaded_ys or {(0, "y_min"): dict(fillcolor="gray", opacity=0.07)}
```
DanielYang59, line 140: @janosh Is this intended or a typo (`shaded_ys = shaded_ys or {(0, "y_min"): dict(fillcolor="gray", opacity=0.07)}`)?

This doesn't conform with the type `shaded_ys: dict[tuple[YMin, YMax], dict[str, Any]] | bool | None`.

Should the key be `(0, "y_max")` (**I didn't look very closely**, just ask)?

tests/test_io.py
```diff
@@ -280,10 +286,10 @@ def mock_urlretrieve(
 
 
 @pytest.mark.parametrize(
-    "compress, font_size, use_styler, width, height",
+    ("compress", "font_size", "use_styler", "width", "height"),
     [
         # Default font size, no compression, DataFrame
-        (False, 14, False, 457, 875),
+        (False, 14, False, 458, 875),
```
DanielYang59, line 292: This is a bit unexpected, I don't think I changed any functional part, but this width changed from `457` to `458`

pymatviz/phonons.py
janosh: thanks for checking! the value is correct here, it's supposed to add a slight gray shading for negative phonon frequencies to visually highlight them (since they indicate dynamic instability of the corresponding structure). but the type is wrong. it should be `dict[tuple[YMin | YMax, YMin | YMax], dict[str, Any]]` since it doesn't matter to plotly whether the first or last value is larger, it's a valid range either way

pyproject.toml
```diff
@@ -119,12 +117,26 @@ isort.split-on-trailing-comma = false
...
-"site/*" = ["D", "INP001", "S602"]
+"__init__.py" = ["F401"]  # unused-import
+"tests/*" = [
+    "D",        # pydocstyle
+    "FBT001",   # boolean-type-hint-positional-argument
+    "FBT002",   # boolean-default-value-positional-argument
+    "INP001",   # implicit-namespace-package
+    "PLR2004",  # magic-value-comparison
+    "S101",     # use of assert
+    ]
```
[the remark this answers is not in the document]
janosh: try installing better-toml VS Code extension. it has good auto-formatting and auto-sorting of arrays on save

.github/workflows/test.yml
```diff
@@ -18,6 +18,7 @@ jobs:
     with:
       os: ${{ matrix.os }}
       python-version: "3.10"
+      test-cmd: "pytest --durations 20 --cov ."
```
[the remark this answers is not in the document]
DanielYang59: Ah I missed this, thanks! Do we want to reduce the default number https://github.com/janosh/workflows/issues/5? Shown all might not be a good default value?

pyproject.toml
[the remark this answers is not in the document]
DanielYang59: Fixed.

> try installing better-toml VS Code extension. it has good auto-formatting and auto-sorting of arrays on save

I do have this installed, but this is not marked for some reason. Fixed.

examples/dataset_exploration/ricci_carrier_transport/convert_dtype+add_strucs.py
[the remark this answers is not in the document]
DanielYang59: renamed!

.github/workflows/test.yml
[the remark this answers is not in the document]
janosh: yes, reducing the default value sounds like a good QoL improvement

pymatviz/phonons.py
DanielYang59: Thanks for the explanation! Would fix this!

janosh (approved):

DanielYang59:
Thanks!

PR #200, opened 2024-09-10
Add `add_annotation` functionality for ptable plotters

### LINKED ISSUE #199: Restore `ptable_hists` kwarg `anno_kwds` (or something similar)
@DanielYang59 re [your comment](https://github.com/janosh/pymatviz/pull/149/files#r1613192588)

> Previously the `ptable_hists` has `symbol_kwargs` and `anno_kwds`. In the new implementation, support for annotate is completely removed, as I would assume their functionality overlaps.

i just realized that the removal of `anno_kwds` caused some issues over in [`matbench-discovery/data/mp/eda_mp_trj.py#L162`](https://github.com/janosh/matbench-discovery/blob/594db5fcde37d8e1dd14371a906bb96f4e41c0a0/data/mp/eda_mp_trj.py#L162) as mentioned in https://github.com/janosh/matbench-discovery/pull/130#discussion_r1733749688. do you remember what you meant by `symbol_kwargs` overlaps with `anno_kwds`. i don't think one can replace the other. this is the output with and without `anno_kwds`. maybe in this case the figure is overplotted with those histogram count annotations but in principle, i think the feature is nice to have (though perhaps with a better name)

Screenshot 2024-09-05 at 16 29 26

DanielYang59:
### Summary

- Add `add_annotation` functionality for ptable plotters, to close #199
    - [x] histogram plotter
    - [x] scatter plotter
    - [x] line plotter
    - [x] heatmap
    - [x] heatmap split
    - [x] heatmap ratio
- Test plotters
    - [x] histogram plotter
    - [x] scatter plotter
    - [x] line plotter
    - [x] heatmap 4f8892b4633de076527d974e98c088777ae6f9f4
    - [x] heatmap split 4f8892b4633de076527d974e98c088777ae6f9f4
    - [x] heatmap ratio 8ed26ac6228f45094b9bd57e95c2066973ff498e
- [x] Add unit tests
- `io.save_fig/save_and_compress_svg` allow non-transparent background

### [Need discussion] different strategy in handling text/annotation color 

`heatmap/heatmap_splits/heatmap_ratio` plotters would accept an additional `anno_text_color` arg to control the annotation text color, while `ptable_{hists/scatters/lines}` would determine the annotation color with the same strategy (`color_elem_strategy`) as the symbol color. This is intended to make the colors less noisy.


assets/ptable-hists.svg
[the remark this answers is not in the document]
DanielYang59: `TODO ++`: expose the `transparent` as an arg otherwise people (or only me) have to hack the source code in order to show the background color: https://github.com/janosh/pymatviz/blob/9c6f9260c1f793d2eef41a7be75291ef31a3cad4/pymatviz/io.py#L134-L139

examples/make_assets/ptable/ptable_matplotlib.py
```diff
@@ -69,7 +69,9 @@
     cbar_axis="x",
     color_elem_strategy="background",
     add_elem_type_legend=True,
-    # x_range=(0, None),
+    symbol_pos=(0.25, 0.8),
+    anno_text={elem.symbol: str(idx - 1) for idx, elem in enumerate(Element)},
```
DanielYang59, line 73: Noticed a funny behavior: 
```python
from pymatgen.core.periodic_table import Element


for idx, ele in enumerate(Element):
    print(idx, ele.symbol)

    if idx > 3:
        break

>>>     0 H
        1 H
        2 H
        3 He
        4 L

print(f"{Element.D.symbol=}")  >>> Element.D.symbol='H'
```

I believe [this is intended to return isotope symbols as its most common one](https://github.com/materialsproject/pymatgen/blob/ea7c3390e2db99a35f3eee242073ffc2830bbd4c/src/pymatgen/core/periodic_table.py#L136-L145).

Then [not sure why define H/D/T separately when they symbol all map to `H`](https://github.com/materialsproject/pymatgen/blob/ea7c3390e2db99a35f3eee242073ffc2830bbd4c/src/pymatgen/core/periodic_table.py#L866-L877).

pymatviz/io.py
```diff
@@ -135,7 +139,7 @@ def save_fig(
     if isinstance(fig, plt.Figure | plt.Axes):
         if hasattr(fig, "figure"):
             fig = fig.figure  # unwrap Axes
-        fig.savefig(path, **kwargs, transparent=True)
+        fig.savefig(path, **kwargs, transparent=transparent_bg)
```
janosh, line 142: need to merge in case `kwargs` already contains `transparent`

```suggestion
        fig.savefig(path, **kwargs | dict(transparent=transparent_bg))
```

pymatviz/ptable/ptable_matplotlib.py
```diff
@@ -70,7 +75,7 @@ def ptable_heatmap(
     cbar_title: str = "Element Count",
     cbar_title_kwargs: dict[str, Any] | None = None,
     cbar_kwargs: dict[str, Any] | None = None,
-    # Migration
+    # Migration (would be removed in the future)
```
janosh, line 78:
```suggestion
    # Migration (will be removed in the future)
```

pymatviz/ptable/ptable_matplotlib.py
```diff
@@ -512,6 +590,15 @@ def ptable_heatmap_splits(
...
     )
 
+    # [Optional] Add annotation
+    if anno_text is not None:
+        projector.add_annotation(
+            text=anno_text or {},
+            pos=anno_pos,
+            text_color=anno_text_color,
+            kwargs=anno_kwargs,
+        )
```
janosh, line 600: how about just passing a single dict to `add_annotation` that includes keys 'text', 'pos', 'color'? otherwise, it can be confusing to remember which things to pass as kwargs and which are separate

pymatviz/ptable/ptable_matplotlib.py
```diff
@@ -698,6 +812,10 @@ def ptable_scatters(
     symbol_text: str | Callable[[Element], str] = lambda elem: elem.symbol,
     symbol_pos: tuple[float, float] = (0.5, 0.8),
     symbol_kwargs: dict[str, Any] | None = None,
+    # Annotation
+    anno_pos: tuple[float, float] = (0.75, 0.75),
+    anno_text: dict[ElemStr, str] | None = None,
+    anno_kwargs: dict[str, Any] | None = None,
```
janosh, line 818: as above, maybe merge `anno_pos` and `anno_text` into `anno_kwargs`

janosh (changes requested):
thanks @DanielYang59! looking good, just some minor suggestions

pymatviz/ptable/ptable_matplotlib.py
DanielYang59: > it can be confusing to remember which things to pass as kwargs and which are separate

I could relate to this. I think the same holds for other methods of projector as well? I would check this.

https://github.com/janosh/pymatviz/blob/8437a9c228ef56cf4475630aa57863cc1f102fed/pymatviz/ptable/_projector.py#L334-L343

pymatviz/ptable/ptable_matplotlib.py
DanielYang59: I'm a little bit unsure about this. For the methods of `PTableProjector`, most args are just passed through from the most outer wrap (`ptable_heatmap` -> `projector.add_annotation()` for example), but for these args, perhaps exposing the most frequently used ones might give users a better idea what could be controlled?

pymatviz/powerups/__init__.py
```diff
@@ -2,6 +2,8 @@
 matplotlib and plotly figures.
 """
 
+from __future__ import annotations
```
janosh, line 5: `pymatviz` is 3.10+. originally, 3.9 was meant to be the last release that needed `from __future__ import annotations` for future type hints to work. i think that was pushed back to 3.12. still not clear to me that we use any annotation features that don't work without it on 3.10 and above. given CI runs fine on 3.10, i think we don't need these

pymatviz/correlation.py
```diff
@@ -43,7 +43,7 @@ def marchenko_pastur_pdf(x: float, gamma: float, sigma: float = 1) -> float:
 
 
 def marchenko_pastur(
-    matrix: ArrayLike,
+    matrix: NDArray,
```
janosh, line 46: i think `ArrayLike` is actually more accurate since `np.linalg.eigvalsh(matrix)` also handles nested lists

pymatviz/powerups/__init__.py
DanielYang59: > originally, 3.9 was meant to be the last release that needed from __future__ import annotations for future type hints to work. i think that was pushed back to 3.12.

[It seems the postponed evaluation is still in TBD state](https://docs.python.org/3/library/__future__.html).

> from __future__ import annotations was previously scheduled to become mandatory in Python 3.10, but the Python Steering Council twice decided to delay the change ([announcement for Python 3.10](https://mail.python.org/archives/list/python-dev@python.org/message/CLVXXPQ2T2LQ5MP2Y53VVQFCXYWQJHKZ/); [announcement for Python 3.11](https://mail.python.org/archives/list/python-dev@python.org/message/VIZEBX5EYMSYIJNDBF6DMUMZOCWHARSO/)). No final decision has been made yet. See also [PEP 563](https://peps.python.org/pep-0563/) and [PEP 649](https://peps.python.org/pep-0649/).

That caused some failure in this PR (I cannot remember which module, I guess its `rdf` in cf8eb60e087b713a96f9782ccec44baf5fb08549), and then I added them for all modules just in case.

janosh: looks like linting and tests on https://github.com/janosh/pymatviz/commit/cf8eb60e087b713a96f9782ccec44baf5fb08549 passed? can we try removing the future imports to see if/where it fails?

janosh: so looks like we do need to the future annotations, at least in `rdf.py`?

DanielYang59: I believe we still need postponed evaluation for most (if not all) of our code, otherwise all type annotations need to be resolved at runtime (i.e. we cannot use `if TYPE_CHECKING` at all)

pymatviz/ptable/ptable_matplotlib.py
DanielYang59: I played around with this and realized passing everything as a single dict doesn't seem to work as I expected, perhaps I misunderstood something?

With current method, the design logic is all "global" args is passed as a separate arg (content of the text, position, color) and all others to the `ax.text` (analogy to a child plotter) as `kwargs`. 

If then pack everything into a single `kwargs` dict (or unpack everything with `**kwargs`), then we still need to annotate inside the method docstring what args are expected and such.

DanielYang59: I must be misunderstanding something, looks like 4d66fe34e494b991382cc2a837022369512a2dd5 doesn't bring much improvement. Because at the end of the day we still need to unpack `kwargs` because the default var name of `plt.text` is not very descriptive inside our ptable plotter (`text(x, y, s, fontdict=None, **kwargs)`) and we still need to map `anno_text` to `s` and such internally.

Perhaps you have a better way around this?

janosh: maybe `dict.setdefault(key1, dict.pop(key2, default))` can help? example: `kwargs.setdefault('t', kwargs.pop('text', ''))`

pymatviz/ptable/ptable_matplotlib.py
janosh: fair point, feel free to mark this as resolved.

pymatviz/powerups/__init__.py
janosh: ok, good to know. let's keep it then

pymatviz/ptable/ptable_matplotlib.py
DanielYang59: Sure thanks!

pymatviz/ptable/ptable_matplotlib.py
DanielYang59: I think this still doesn't bypass the need to map `t` to `text` inside the `add_annotation` method, also this might make debugging hard as user (or us) wouldn't know what the value really come from (`t` or `text`).

DanielYang59: Sorry I re-read that code again and apparently I misunderstood. That should map `key2` to `key1`.

DanielYang59: Sorry I don't think I'm really following, do you mean something like f10bebcb4cdd0d2648ba0a6de0e8508375a7b481? Or if not perhaps can you perhaps push a commit to get me on board?

janosh: see [365a296](https://github.com/janosh/pymatviz/pull/200/commits/365a29661735f3e9cfa1392a3c3bc1dbecb08b04). if you don't think it's an improvement. feel free to ignore this entire discussion and go back to what you had originally

pymatviz/ptable/_projector.py
```diff
@@ -552,8 +550,8 @@ def add_annotation(
                 symbol_color = text_color
 
             ax.text(
-                *pos,
-                text.get(symbol, ""),
+                *kwargs.pop("pos"),
```
DanielYang59, line 553: As this operates inside a loop, I guess this would only get the desired value for the first iteration?

pymatviz/ptable/ptable_matplotlib.py
DanielYang59: > if you don't think it's an improvement.

This sounds like a harsh comment for me to make. But well...I would prefer personally to revert to what we had originally at this moment, and maybe go back to this if we came up with a better solution (as this is internal and wouldn't impact the public API much), which I don't have right now.

---

With current implementation, pros:
- The interface to the projector method looks cleaner and we don't need to remember what to pass as separate args. (But the IDE would hint and we don't need to remember.)

Cons:
- We still need to handle some keywords separately (`text_color`, `text`) inside the method, as not args are passed to `ax.text` directly.
- We lost the type annotation, and partly the default values (I prefer if we don't need to read the source code to know the default values, otherwise IDE would also show the default values).

janosh: good points. then let's revert to your earlier implementation

DanielYang59: This branch should be in the original state if I didn't miss anything, perhaps you could review this again? Thanks!

janosh (approved):
thanks a lot for this @DanielYang59! 👍

DanielYang59:
Cool! Thanks a lot for the final touch and review.

PR #207, opened 2024-09-25
Clean up var names in unit tests, avoid MP API access in GitHub workflow

DanielYang59:
### Summary

- Clean up var names in unit tests, make module level var capital and remove a few unused vars
- Avoid fetching structure from MP database as API key cannot be passed to actions triggered by forks


DanielYang59:
I believe I have seen this error in another PR or somewhere:
```
File "/opt/hostedtoolcache/Python/3.10.15/x64/lib/python3.10/site-packages/mp_api/client/core/client.py", line 1031, in _submit_request_and_process
    raise MPRestError(
mp_api.client.core.client.MPRestError: REST query returned with error status code 401 on URL https://api.materialsproject.org/materials/core/?deprecated=False&_fields=structure&material_ids=mp-19017&_limit=1000 with message:
Response {
  "message":"No API key found in request"
}
Error: Process completed with exit code 1.
```

Looks like MP API key got lost sometimes?

janosh:
i can't reproduce the error locally. just tried with the latest version `mp-api==0.42.2`. the repo definitely has the secret

```
gh secret list
NAME        UPDATED          
MP_API_KEY  about 24 days ago
PYPI_TOKEN  about 1 year ago
```

and it's passed correctly

https://github.com/janosh/pymatviz/blob/7e32978b4fe1051bf58fa174c122b781cb0ff2f9/.github/workflows/test.yml#L59

DanielYang59:
Yep that occurs intermittently, not sure what the reason is

DanielYang59:
Interestingly I could recreate that MP_API_KEY not found issue reliably.

Do you think this is a bug or I wasn't using that key correctly? @janosh If the former I would go and report this.

```python
import os
from mp_api.client import MPRester


os.environ["MP_API_KEY"] = "not shown here"

mp_api = os.environ.get("MP_API_KEY")
assert mp_api is not None

struct = MPRester().get_structure_by_material_id("mp-19017")
```

```
  File "z:\test\test_mp_api.py", line 10, in <module>
    struct = MPRester().get_structure_by_material_id("mp-19017")
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "Z:\test\venv\Lib\site-packages\mp_api\client\mprester.py", line 462, in get_structure_by_material_id
    structure_data = self.materials.get_structure_by_material_id(
                     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "Z:\test\venv\Lib\site-packages\mp_api\client\routes\materials\materials.py", line 124, in get_structure_by_material_id
    response = self.search(material_ids=material_id, fields=[field])
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "Z:\test\venv\Lib\site-packages\mp_api\client\routes\materials\materials.py", line 245, in search
    return super()._search(
           ^^^^^^^^^^^^^^^^
  File "Z:\test\venv\Lib\site-packages\mp_api\client\core\client.py", line 1181, in _search
    return self._get_all_documents(
           ^^^^^^^^^^^^^^^^^^^^^^^^
  File "Z:\test\venv\Lib\site-packages\mp_api\client\core\client.py", line 1254, in _get_all_documents
    results = self._query_resource(
              ^^^^^^^^^^^^^^^^^^^^^
  File "Z:\test\venv\Lib\site-packages\mp_api\client\core\client.py", line 565, in _query_resource
    data = self._submit_requests(
           ^^^^^^^^^^^^^^^^^^^^^^
  File "Z:\test\venv\Lib\site-packages\mp_api\client\core\client.py", line 712, in _submit_requests
    initial_data_tuples = self._multi_thread(
                          ^^^^^^^^^^^^^^^^^^^
  File "Z:\test\venv\Lib\site-packages\mp_api\client\core\client.py", line 934, in _multi_thread
    data, subtotal = future.result()
                     ^^^^^^^^^^^^^^^
  File "C:\Users\yang\anaconda3\Lib\concurrent\futures\_base.py", line 449, in result
    return self.__get_result()
           ^^^^^^^^^^^^^^^^^^^
  File "C:\Users\yang\anaconda3\Lib\concurrent\futures\_base.py", line 401, in __get_result
    raise self._exception
  File "C:\Users\yang\anaconda3\Lib\concurrent\futures\thread.py", line 58, in run
    result = self.fn(*self.args, **self.kwargs)
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "Z:\test\venv\Lib\site-packages\mp_api\client\core\client.py", line 1027, in _submit_request_and_process
    raise MPRestError(
mp_api.client.core.client.MPRestError: REST query returned with error status code 401 on URL https://api.materialsproject.org/materials/core/?deprecated=False&_fields=structure&material_ids=mp-19017&_limit=1000 with message:
Response {
  "message":"No API key found in request"
}
```

janosh:
@DanielYang59 thanks for debugging. agreed that `MP_API_KEY` should not be read at module import time but during `MPRester` instance creation. pinging @tsmathis and @esoteric-ephemera in case you want to consider changing

not sure that's the root cause here though. `MP_API_KEY` should be available before python is invoked

```yml
- name: Run script
  run: python ${{ matrix.script }}
  env:
    MP_API_KEY: ${{ secrets.MP_API_KEY }}
```

DanielYang59:
> agreed that MP_API_KEY should not be read at module import time but during MPRester instance creation.

I might open an issue there for easier tracking

> not sure that's the root cause here though. MP_API_KEY should be available before python is invoked

100% agree, I don't think that's the issue (though not sure how and when GitHub runner injected env variables). I inserted an `assert mp_api_key` is not None before calling `MPRester` but it seems to pass? I'm super confused at this moment :(

examples/make_assets/structure_viz.py
```diff
@@ -30,6 +30,9 @@
 
 
 # %% Plot some disordered structures in 2D
+if os.environ.get("MP_API_KEY") is None:
```
DanielYang59, line 33: > not sure that's the root cause here though. MP_API_KEY should be available before python is invoked

100% agree, I don't think that's the issue (though not sure how and when GitHub runner injected env variables). I inserted an `assert mp_api_key is not None` before calling `MPRester` and it seems to pass? I'm super confused at this moment :(

DanielYang59:
@janosh  can you have another look at the API_KEY secret (I don't have access to that)? It looks like being empty?
> RuntimeError: mp api key is empty

examples/make_assets/structure_viz.py
```diff
@@ -31,11 +30,20 @@
 
 
 # %% Plot some disordered structures in 2D
+mp_api_key = os.environ.get("MP_API_KEY")
+if mp_api_key is None:
+    raise RuntimeError("api key is None")
+
+elif mp_api_key is not None and len(mp_api_key.strip()) == 0:
+    raise RuntimeError("mp api key is empty")
```
DanielYang59, line 38: @janosh  can you have another look at the API_KEY secret (I don't have access to that)? [It looks like being empty](https://github.com/janosh/pymatviz/actions/runs/11071283116/job/30762788796?pr=207)?
> RuntimeError: mp api key is empty

janosh: i can't see it either. once you set a secret you can only change it's value, not see the current value. i just reset the secret to the what i think it was before so let's see if that makes a difference

DanielYang59: Still no luck. I'm not seeing the `test.yml` or the template from your `workflows` being changed at all recently, so super confused what is causing this failure.

 I would have another look tomorrow because I don't have my laptop with me ATM (which I usually use to code), so everything feels very unnatural :)

Another thread I haven't got to look at, not sure if it's related at all: [ Github Actions reusable workflow doesn't recognize secrets #69082 ](https://github.com/orgs/community/discussions/69082)

janosh: > not sure if it's related at all

don't think so, the `test-scripts` job doesn't have a reusable workflow

DanielYang59: I believe it's related to [secrets access from forked repos](https://docs.github.com/en/actions/security-for-github-actions/security-guides/using-secrets-in-github-actions?tool=webui):
> With the exception of GITHUB_TOKEN, secrets are not passed to the runner when a workflow is triggered from a forked repository.

As you have noticed both this PR and #210 were created from our forks.

However above documentation doesn't provide any workaround for this :(

janosh: that sounds like it could be the culprit. probably the thing to do is to rather than fetch those structures on the fly, save them as JSON to `tests/files/structs` and load them from there in this script (while leaving the code that generated those structures files in place but behind `if no os.path.isfile(...)` as documentation)

DanielYang59: You always have helpful suggestions :) How is 95614f26fd63f0213b8cd856bff9cae6b06679b1 looking? I don't really like to manually download and save files or such, it's always better to automate this

DanielYang59:
Notice another potential issue, both [our test suite](https://github.com/janosh/pymatviz/actions/runs/11089181077/job/30809986835) and [`pre-commit`](https://results.pre-commit.ci/run/github/340898532/1727579206.RhGO6GI2RnWbVLy7mwvnSg) seems to run on main branch for some reason.

Note at commit 1f11e0b5f1c0371354a3ece9e2708753b3cf2763, main branch has NOT been merged and [`test_structure_viz.py` has a total of 224 lines](https://github.com/janosh/pymatviz/blob/1f11e0b5f1c0371354a3ece9e2708753b3cf2763/tests/test_structure_viz.py) while errors are reported on recently included lines that hasn't been merged.

examples/make_assets/structure_viz.py
```diff
@@ -30,22 +31,19 @@
...
-    for mp_id in ["mp-19017", "mp-12712"]
-}
-
-struct: Structure  # for type hinting
-
-for mp_id, struct in disordered_structs.items():
+struct_mp_ids = ("mp-19017", "mp-12712")
+structure_dir = f"{TEST_FILES}/structure"
+
+if not os.environ.get("GITHUB_ACTIONS"):
```
janosh, line 37: `if not os.environ.get("CI"):` would be more generic and compatible with other CI providers. but i think the check shouldn't be for CI at all but rather if the structure files already exist.

tests/files/structures/mp-12712.json
janosh, line 1: let's make these structure files `.json.gz`. we may want to add more structures in the future so good to reduce file size and line diff counts

janosh (changes requested):
thanks a lot @DanielYang59! much clearer now, which vars are module-scoped 👍

examples/make_assets/structure_viz.py
DanielYang59: Brilliant!

examples/make_assets/structure_viz.py
```diff
@@ -34,21 +34,22 @@
...
 structure_dir = f"{TEST_FILES}/structures"
 
-if not os.environ.get("GITHUB_ACTIONS"):
-    os.makedirs(structure_dir, exist_ok=True)
-    for mp_id in struct_mp_ids:
+os.makedirs(structure_dir, exist_ok=True)
+for mp_id in struct_mp_ids:
+    struct_file = f"{structure_dir}/{mp_id}.json.gz"
+    if not os.path.exists(struct_file):
+        if os.environ.get("CI"):
```
janosh, line 41: maybe use `os.getenv`

examples/make_assets/structure_viz.py
```diff
@@ -34,21 +34,22 @@
 struct_mp_ids = ("mp-19017", "mp-12712")
 structure_dir = f"{TEST_FILES}/structures"
 
-if not os.environ.get("GITHUB_ACTIONS"):
-    os.makedirs(structure_dir, exist_ok=True)
-    for mp_id in struct_mp_ids:
+os.makedirs(structure_dir, exist_ok=True)
+for mp_id in struct_mp_ids:
+    struct_file = f"{structure_dir}/{mp_id}.json.gz"
+    if not os.path.exists(struct_file):
```
janosh, line 40: `os.path.isfile` is slightly more precise and semantic

janosh (changes requested):

examples/make_assets/structure_viz.py
DanielYang59: Ah, must be a typo, should have used it

janosh (approved):
thanks for uncovering https://github.com/actions/checkout/issues/1359. didn't know about this subtlety. will try to remember if it comes up again

DanielYang59:
No problem! Thanks for reviewing (always amazed how much you know about all those APIs)

janosh:
maybe time for a new release soon? what do you think? anything that should be done first?

DanielYang59:
Perhaps let me finish #200 perhaps in one or two days? It's done, I just want to finish the interactive tester and have a try on #200?

janosh:
sure no hurry, we can definitely wait 2 days or more. looking forward to #200! 👍

DanielYang59:
I really appreciate that :)

Issue #22, opened 2022-02-28
Add image comparison tests for all `matplotlib` plots


janosh:
As in #20 (`tests/test_struct_vis.py`), use `matplotlib.testing.compare.compare_images()` to test all functions that generate `matplotlib` figures.

janosh:
Maybe actually not worth the trouble. Makes tests very brittle to the point where the image diff tolerance needs to be so high that it seems less valuable. [2882acf](https://github.com/janosh/pymatviz/pull/28/commits/2882acf2f03d1eb31ce8e49dc3bacceabb2bb289) where diff RMS was bumped from 0.8 to 100/

PR #208, opened 2024-09-25
Allow interactively inspect plots

DanielYang59:
### Summary

Allow interactively inspect plots.

**Objective:**
 Add an interactive inspection tool to prompt the user to inspect a plot, this would only run locally and skip CI workflow.

**Rationale:** 
It's really hard to automate tests for plotters, we could assert some elements (title, number of subplots and such), but it's hard to ensure the overall figure is in expected shape. Therefore we might be better off insert some interactive checkers which would only run locally anyway. Such that if developers want to inspect some plotters, they could run that particular unit test locally.

Currently I mostly use the "make_assets" scripts to inspect plots, but it's largely limited to certain "styles". Also it's not quite reusable if I wish to inspect some other styles.


janosh:
the way this is usually solved is with reference files and doing pixel-averaged distance between those and newly generated plots. see [`matplotlib.testing.compare.compare_images`](https://matplotlib.org/stable/api/testing_api.html#matplotlib.testing.compare.compare_images)

i briefly looked into doing that in https://github.com/janosh/pymatviz/issues/22 but seemed too brittle to pursue. also a bit annoying having to have a lot of references files in `git` history, esp. since they i suspect they need to be bitmapped. could maybe use the SVG assets and convert them to PNGs on the fly but that would slow down tests.

DanielYang59:
I guess my option is more naive: "just ask the user to inspect the figures themselve to see if it's broken", with a prompt like `please check if the colorbar is displayed properly`. 

I intend to make this optional on top of the regular test workflow (only trigger in non-CI environment), such that if we did some major change to the code base, we run these tests locally to see if there is any major breakage.

I would implement first and see how everything goes :)

DanielYang59:
I would close this do merge this into #200 as that's a better place to test this function out.

janosh:
is this PR still on the to do list or could it be closed?

DanielYang59:
> is this PR still on the to do list or could it be closed?

Sorry for the lack of response and progress, I cleared most of my TODO list from the MP side so I would get my hands on these by next week ;)

DanielYang59:
I think I'm happy with current functionality so far, it displays the figure optionally when invoked interactively, and doesn't require much (if any) code change in our current test code.

However there may be a few quality-of-life improvement possible:
- Current implementation saves the figure to a temporary location and re-open it (otherwise the display style might change with window size and so on), this sounds a bit stupid, wondering if there's a way to display the figure directly and show the same result as saved figure?
- `pytest` would capture output by default, and to use this we have to explicitly `pytest -s test_script`, I tried but didn't manage to release the capture from inside the `interactive_check` helper function. 
- displayed figures tend to be larger than I like it to be, I tried to scale it by monitor size (say 25% of the screen and at top-center) but didn't have much luck so far. I previously found a SO thread (forgot which one now) saying the API to get display size is OS-dependent, but I would have another try

DanielYang59:
Just I noticed [another `pytest` plugin from `matplotlib` to compare images by their RMS](https://github.com/matplotlib/pytest-mpl):
> pytest-mpl is a [pytest](https://docs.pytest.org/) plugin to facilitate image comparison for [Matplotlib](http://www.matplotlib.org/) figures.

> For each figure to test, an image is generated and then subtracted from an existing reference image. If the RMS of the residual is larger than a user-specified tolerance, the test will fail. Alternatively, the generated image can be hashed and compared to an expected value.

> For more information, see the [pytest-mpl documentation](https://pytest-mpl.readthedocs.io/).

pyproject.toml
```diff
@@ -70,7 +70,7 @@ build-backend = "setuptools.build_meta"
 
 [tool.pytest.ini_options]
 testpaths = ["tests"]
-addopts = "-p no:warnings"
+addopts = "-s -p no:warnings"
```
DanielYang59, line 73: Admittedly this is not the best option (to globally un-capture output) as I was hoping this could be integrated without any code change to our current test framework, but I didn't find another way around this issue.

The most related might be the [`capsys` fixture](https://docs.pytest.org/en/stable/how-to/capture-stdout-stderr.html#accessing-captured-output-from-a-test-function) but again I didn't manage to integrate this fixture into this helper function.

@janosh  Not sure if you would be happy with adding this option to globally release `pytest` output capture (though I would assume the side effect to be limited as we aren't inserting much `print` and so on)
 
Also for the other possible improvements mentioned in https://github.com/janosh/pymatviz/pull/208#issuecomment-2558066383, unfortunately I didn't find better solutions. Not sure if you have any comment?

Happy coding 2025 in any case :)

DanielYang59:
I might just close this PR now that pmv is actively migrating to `plotly` :)

Issue #216, opened 2024-09-30
Unable to install with extras={test,export-figs,make-assets} on mac due to upstream issue with kaleido


theochemtheo:
Hello,

I had trouble installing `pymatviz` when setting up `matbench-discover` today on my Mac due to the dependency on `kaleido>=0.2` (thanks for both packages btw).

I believe this is due to an upstream mistake with the most recent release of `kaleido`, as detailed in [this issue](https://github.com/plotly/Kaleido/issues/176). In brief, they [didn't include wheels](https://github.com/plotly/Kaleido/releases/tag/v0.2.1.post1) for non-linux platforms in their most recent release (`0.2.1post1`).

I was able to fix this in my own fork by pinning to `kaleido=0.2.1`, which hopefully will only be necessary in the short term. I will open a PR including this fix, if you're open to that.

DanielYang59:
@theochemtheo thanks for reporting this, I'm being very curious as I cannot reproduce this on my MacOS machine. May I know what command you used to install pymatviz with extras? `pip` doesn't seem to pick that version as neither source distribution nor compatible binary distribution is available, perhaps you are using a python package manager other than pip?

DanielYang59:
Thanks for the input! I could recreate this issue with `uv`:
```
>>> uv pip install -r pyproject.toml --extra test
# be sure to remove the "uv.lock" cache file
>>> uv sync --extra test
error: distribution kaleido==0.2.1.post1 @ registry+https://pypi.org/simple can't be installed because it doesn't have a source distribution or wheel for the current platform
```

I might look deeper into this because it seems the resolver for `uv` is not working property.

PR #217, opened 2024-09-30
pin kaleido==0.2.1

theochemtheo:
* Pins `kaleido==0.2.1`, closing #216

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -40,15 +40,15 @@ test = [
     "adjustText>=0.8",
     "bs4>=0.0.2",
     "cssutils>=2.9",
-    "kaleido>=0.2",
+    "kaleido==0.2.1",
     "lxml>=4.9",
     "pdfCropMargins>=2",
     "pytest-cov>=5",
     "pytest>=8",
     "weasyprint>=60",
 ]
-make-assets = ["kaleido>=0.2", "matminer>=0.9", "mp_api>=0.42"]
-export-figs = ["kaleido>=0.2"]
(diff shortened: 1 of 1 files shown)


janosh (approved):
@theochemtheo thanks for reporting and fixing! 👍

DanielYang59:
Should that pin be [`0.2.1.post1`](https://github.com/plotly/Kaleido/releases) instead of `0.2.1post1`?

For some reason I didn't see `kaleido==0.2.1.post1` in my available versions from `pip` for some reason (I'm also working on MacOS)?
```
>>> pip index versions kaleido
kaleido (0.2.1)
Available versions: 0.2.1, 0.2.0
```

janosh:
[yes, good catch](https://pypi.org/project/kaleido/0.2.1.post1/)

DanielYang59:
Can you install `pip install kaleido==0.2.1.post1` (with dot before `post1`)? I cannot install it either:
```
pip install kaleido==0.2.1.post1
Looking in indexes: https://pypi.tuna.tsinghua.edu.cn/simple
ERROR: Could not find a version that satisfies the requirement kaleido==0.2.1.post1 (from versions: 0.2.0rc1, 0.2.0, 0.2.1)
ERROR: No matching distribution found for kaleido==0.2.1.post1
```

janosh:
get the same error but that's consistent with there being [only a linux wheel](https://pypi.org/project/kaleido/0.2.1.post1/#files)

DanielYang59:
Yes there's only a wheel for ARM32 Linux (I don't think pip would automatically pick this version) so I was wondering why the original author would have that issue (apple silicon is ARM64).

PR #218, opened 2024-10-01
`structure_(2|3)d_plotly` allow overriding subplot title's y position and anchor   

janosh:
includes new test


janosh:
also includes a new unrelated asset. heatmap of VASP valence electron counts

ptable-heatmap-plotly-vasp-psp-n-valence-electrons

PR #226, opened 2024-10-08
Fix `ptable_heatmap_ratio` following change of default `count_elements` `fill_value` to 0

### LINKED ISSUE #215: behavior change in `ptable_heatmap_ratio`
Unexpected behavior change in `ptable_heatmap_ratio`, some blue tiles (not in 2nd tile) now becomes grey (not in 1nd tile)

ptable-heatmap-ratio

`0.10.0` seems to work correctly, but `0.10.1` is broken.

DanielYang59:
### Summary

- Change default `fill_value` of `count_elements` from `0` to `None`, to close #215 and https://github.com/janosh/pymatviz/pull/192#discussion_r1789013965
- [x] Check example ipynb (mostly `ptable_heatmap`)
- [x] add unit test

####  `ptable_heatmap`

- `value_mode` as `log`: no impact as near-zero values are filtered
- `value_mode` as `fraction/percent`: **buggy behavour, would fix in #230**

#### `histogram.elements_hist`

- no impact owing to "elem_count > 0" filter

#### `ptable_heatmap_plotly`

- no impact as `fill_value` is passed from `ptable_heatmap_plotly`


pymatviz/histogram.py
```diff
@@ -74,7 +74,7 @@ def elements_hist(
     """
     ax = ax or plt.gca()
 
-    elem_counts = count_elements(formulas, count_mode)
+    elem_counts = count_elements(formulas, count_mode, fill_value=0)
     non_zero = elem_counts[elem_counts > 0].sort_values(ascending=False)
```
DanielYang59, line 78: tag myself: If pandas series could automatically filter out None, this could be reverted.

tests/ptable/test_ptable_matplotlib.py
```diff
@@ -22,7 +22,7 @@
 
 @pytest.fixture
 def glass_elem_counts(glass_formulas: pd.Series[Composition]) -> pd.Series[int]:
-    return pmv.count_elements(glass_formulas)
+    return pmv.count_elements(glass_formulas, fill_value=0)
```
DanielYang59, line 25: This is added to better match the type annotation (`pd.Series[int]`), for a "counting" task, using zero for missing element may be more appropriate.

pymatviz/histogram.py
janosh: not sure if this is still relevant but isn't that what `pd.Series.dropna()` is for?

pymatviz/process_data.py
```diff
@@ -50,13 +52,20 @@ def count_elements(
...
-        fill_value (float | None): Value to fill in for missing elements. Defaults to 0.
+        exclude_elements (Sequence[str]): Elements to exclude from the count.
+            Defaults to ().
+        fill_value (float | None): Value to fill in for missing elements.
+            Defaults to None for NaN.
 
     Returns:
         pd.Series: Map element symbols to heatmap values.
     """
+    # TODO: remove warning after 2025-10-08
```
janosh, line 63: i think keeping this warning for 2 months till 2025-01-01 should be enough

examples/mp_bimodal_e_form.ipynb
janosh, line 1: would be good to set `pymatviz_white` template in this notebook

janosh (approved):
looks good. 👍 

the PR title is slightly misleading as the default value `fill_value` of `count_elements` is already `None`. maybe better "fix `ptable_heatmap_ratio` following change of default `count_elements` `fill_value` to 0"?

DanielYang59:
> the PR title is slightly misleading as the default value fill_value of count_elements is already None

It was intended as a late warning, as the original breaking change is kind of silent. But yep I'm happy with the new title

pymatviz/histogram.py
DanielYang59: Yep I forgot to resolve this comment because I wasn't too sure about the behaviour of `elem_counts > 0`, and it turns out it would filter out None.

PR #229, opened 2024-10-09
Delete `correlation.py` module

janosh:
Arguably plotting Marchenko–Pastur distros is too specialized and narrow-purpose to remain in the package


DanielYang59:
Hi @janosh I appreciate all your work, but can we perhaps follow the best practice and at least emit a deprecation warning for some period before deletion? Even though it’s quite specialized, it might still be in use by someone?

janosh:
we could but i'm close to certain nobody ever used this function except me. i only used it in one project couple years ago and i never needed it again after that. that said, i'm open to a deprecation period.

DanielYang59:
I would trust your judgement here until someone complains (hopefully not) :)

PR #231, opened 2024-10-10
Reduce duplicate in dependency declare

DanielYang59:
### Summary

- Reduce duplicate in dependency declare, new trick picked up from https://github.com/hackingmaterials/matminer/pull/950#discussion_r1792190959


janosh (approved):
thanks! i was also planning on doing that 😄 👍

PR #238, opened 2024-10-20
Speedup import and add regression check for import time

### LINKED ISSUE #209: `pymatviz` import cost way too high
takes almost 2 seconds to import the latest version of `pymatviz`. used to be a lot less. i suspect #194 may have caused this problem without me noticing at the time. current import waterfall can be generated with:

```
time python -X importtime -c 'import pymatviz'
...
fraction
import time:       157 |        157 |       pymatgen.analysis.diffraction.core
import time:       624 |        966 |     pymatgen.analysis.diffraction.xrd
import time:       276 |       1241 |   pymatviz.xrd
import time:       175 |        175 |   pymatviz.histogram
import time:       992 |    2308598 | pymatviz
python -X importtime -c 'import pymatviz'  1.88s user 0.33s system 82% cpu 2.705 total
```

should figure out which modules take the longest to load, and decide if they really need to be imported from `pymatviz/__init__.py` or if they could be lazy-loaded instead (i.e. only loaded if the user actually accesses them) and put some regression tests in place to ensure no future PRs blow up the import time again. might also make sense to see how large code bases like pandas keep their import time low

DanielYang59:
### Summary

- Speedup import, to close #209
- [x] Add check for import time
- [ ] Update `.pytest-split-durations` (how to do this?)

Profile command: `python -X importtime -c "import pymatviz" 2> pmv.log && tuna pmv.log`

### DIFF
--- a/pymatviz/structure_viz/plotly.py
+++ b/pymatviz/structure_viz/plotly.py
@@ -4,11 +4,9 @@ from __future__ import annotations
 
 import math
 import warnings
-from collections.abc import Callable, Sequence
-from typing import TYPE_CHECKING, Any, Literal
+from typing import TYPE_CHECKING
 
 import numpy as np
-import plotly.graph_objects as go
 from plotly.subplots import make_subplots
 from pymatgen.analysis.local_env import CrystalNN, NearNeighbors
 from pymatgen.symmetry.analyzer import SpacegroupAnalyzer
--- a/pymatviz/utils.py
+++ b/pymatviz/utils.py
@@ -19,7 +19,6 @@ import plotly.io as pio
 from matplotlib.colors import to_rgb
 from matplotlib.offsetbox import AnchoredText
 from matplotlib.ticker import FormatStrFormatter, PercentFormatter, ScalarFormatter
-from pymatgen.core import Structure
 
 
 if TYPE_CHECKING:
@@ -680,7 +679,7 @@ def _get_matplotlib_font_color(fig: plt.Figure | plt.Axes) -> str:
 
 def normalize_to_dict(
     inputs: T | Sequence[T] | dict[str, T],
-    cls: type[T] = Structure,
+    cls: type[T] | None = None,
     key_gen: Callable[[T], str] = lambda obj: getattr(
         obj, "formula", type(obj).__name__
     ),
@@ -700,8 +699,14 @@ def normalize_to_dict(
     Raises:
         TypeError: If the input format is invalid.
     """
+    if cls is None:
+        from pymatgen.core import Structure  # costly import
+
+        cls = Structure
+
     if isinstance(inputs, cls):
-        return {"": inputs}
+        return {"": inputs}  # type: ignore[dict-item]
+
     if (
         isinstance(inputs, list | tuple)
         and all(isinstance(obj, cls) for obj in inputs)
--- a/pymatviz/utils.py
+++ b/pymatviz/utils.py
@@ -16,9 +16,11 @@ import numpy as np
 import pandas as pd
 import plotly.graph_objects as go
 import plotly.io as pio
+import scipy.stats
 from matplotlib.colors import to_rgb
 from matplotlib.offsetbox import AnchoredText
 from matplotlib.ticker import FormatStrFormatter, PercentFormatter, ScalarFormatter
+from pymatgen.core import Structure
 
 
 if TYPE_CHECKING:
@@ -233,8 +235,6 @@ def bin_df_cols(
         )
 
     if density_col:
-        import scipy.stats  # expensive import
-
         # compute kernel density estimate for each bin
         values = df_in[bin_by_cols].dropna().T
         gaussian_kde = scipy.stats.gaussian_kde(values.astype(float))
@@ -679,7 +679,7 @@ def _get_matplotlib_font_color(fig: plt.Figure | plt.Axes) -> str:
 
 def normalize_to_dict(
     inputs: T | Sequence[T] | dict[str, T],
-    cls: type[T] | None = None,
+    cls: type[T] = Structure,
     key_gen: Callable[[T], str] = lambda obj: getattr(
         obj, "formula", type(obj).__name__
     ),
@@ -699,13 +699,8 @@ def normalize_to_dict(
     Raises:
         TypeError: If the input format is invalid.
     """
-    if cls is None:
-        from pymatgen.core import Structure  # costly import
-
-        cls = Structure
-
     if isinstance(inputs, cls):
-        return {"": inputs}  # type: ignore[dict-item]
+        return {"": inputs}
 
     if (
         isinstance(inputs, list | tuple)
@@ -728,23 +723,3 @@ def normalize_to_dict(
     raise TypeError(
         f"Invalid {inputs=}, expected {cls_name} or dict/list/tuple of {cls_name}"
     )
-
-
-def _check_type(obj: object, type_str: tuple[str, ...] | str) -> bool:
-    """Alternative to isinstance that avoids imports.
-
-    Todo:
-    Taken from monty.json, use until monty.json import fix merged.
-
-    Note for future developers: the type_str is not always obvious for an
-    object. For example, pandas.DataFrame is actually pandas.core.frame.DataFrame.
-    To find out the type_str for an object, run type(obj).mro(). This will
-    list all the types that an object can resolve to in order of generality
-    (all objects have the builtins.object as the last one).
-    """
-    type_str = type_str if isinstance(type_str, tuple) else (type_str,)
-    try:
-        mro = type(obj).mro()
-    except TypeError:
-        return False
-    return any(f"{o.__module__}.{o.__name__}" == ts for o in mro for ts in type_str)
--- a/pymatviz/utils.py
+++ b/pymatviz/utils.py
@@ -16,7 +16,6 @@ import numpy as np
 import pandas as pd
 import plotly.graph_objects as go
 import plotly.io as pio
-import scipy.stats
 from matplotlib.colors import to_rgb
 from matplotlib.offsetbox import AnchoredText
 from matplotlib.ticker import FormatStrFormatter, PercentFormatter, ScalarFormatter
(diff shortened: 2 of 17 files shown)


pymatviz/utils.py
```diff
@@ -699,8 +699,14 @@ def normalize_to_dict(
     Raises:
         TypeError: If the input format is invalid.
     """
+    if cls is None:
+        from pymatgen.core import Structure  # costly import
+
+        cls = Structure
+
     if isinstance(inputs, cls):
```
DanielYang59, line 707: Looks like we might need a temporary helper function to test Structure until all those changes go to pymatgen (would certainly take some time)

DanielYang59:
I have a bad feeling about this PR, I think I might prefer to wait until the import fix in monty get merged:
- It's badly polluting the code base by lazily almost everything about pymatgen, because any import involving `from monty.json import xxx` could potentially slow down the import (if `torch` is installed).

---

Meanwhile for some reason every time I lazily import one package, there seems to be another package popping up from nowhere and make the total import time roughly the same or even more.

For example if I profile on main branch (without torch):
image

And the tip of this branch (notice sklearn appeared for some reason):

image

janosh:
Thanks a lot for digging into this, really appreciate that! I suspect you're right, we're dependent on pull requests in upstream repositories being merged to achieve significant improvements in pumatviz.

pymatviz/structure_viz/plotly.py
```diff
@@ -4,11 +4,9 @@
 
 import math
 import warnings
-from collections.abc import Callable, Sequence
-from typing import TYPE_CHECKING, Any, Literal
```
DanielYang59, line 8: No problem at all :)

Putting aside the import time, it appears our import linter is not doing its job?
- Some completely unused imports are not removed, like this one.
- Some type checking imports are not relocated.
- Even duplicate imports are not removed (import both at runtime and for type checking)

janosh: didn't encounter `ruff` import sorting issues yet. is it still happening? can you give a specific example?

DanielYang59:
I will do this :) I was planning to add a runtime unit test as well (perhaps to core parts of pymatgen as well), now we only test the results, and runtime is not covered at all

pymatviz/structure_viz/plotly.py
DanielYang59: Yes I'm not seeing `ruff` errors from main branch either (when there should):
```
(venv) yang@Yang-MacBook pymatviz % ruff --version
ruff 0.7.0
(venv) yang@Yang-MacBook pymatviz % ruff check . --select I001
All checks passed!
(venv) yang@Yang-MacBook pymatviz % git branch
  count-elems-default-fill-val
* main
```

- Some completely unused imports are not removed
I must have misread here, they should be duplicate imports

- Some type checking only imports are not relocated
Check the `pymatviz/coordination.py` change in this PR, `ruff` doesn't report an error even with `from __future__ import annotations`:

image

- Even duplicate imports are not removed (import both at runtime and for type checking)

It's everywhere in this PR (check the rest of the diff)

DanielYang59:
Regarding the actual test methodology, I haven't got much finding about doing a clean and uncached import. Your method currently works, what about spawning a subprocess (is there any pitfall about this?), like:
```python
import subprocess
import time


def measure_import_time(module_name):
    start_time = time.time()
    subprocess.run(['python', '-c', 'import pymatviz'], check=True)
    return time.time() - start_time
```

Is there any advantage to use relative time (`time.perf_counter()`) instead of real-world time (`time.time()`)? The latter seems more comprehensible to me (though at the end of the day we still need to collect the reference time from CI).

janosh:
> Your method certainly works, what about spawning a subprocess (is there any pitfall about this, like subprocess overhead maybe?), something like:

i'm not sure but i think `subprocess` should be fine since afaik it runs in a completely isolated process with separate import caching


> Is there any advantage to use [relative time (`time.perf_counter()`)](https://docs.python.org/3/library/time.html#time.perf_counter) over real-world time (`time.time()`)?

yes, [have a look at this video](https://www.youtube.com/watch?v=hb-12mgQrzM) which explains the virtues of `time.perf_counter`

DanielYang59:
> > Your method certainly works, what about spawning a subprocess (is there any pitfall about this, like subprocess overhead maybe?), something like:
> 
> i'm not sure but i think `subprocess` should be fine since afaik it runs in a completely isolated process with separate import caching

Thanks! Let's use subprocess for now, this should make things much easier (if there is not pitfall)

> > Is there any advantage to use [relative time (`time.perf_counter()`)](https://docs.python.org/3/library/time.html#time.perf_counter) over real-world time (`time.time()`)?
> 
> yes, [have a look at this video](https://www.youtube.com/watch?v=hb-12mgQrzM) which explains the virtues of `time.perf_counter`

Thanks for sharing, I would have a look later but I certain trust you in this regard.

By the way, how do we regenerate the `.pytest-split-durations` record (not now, after we finish everything, current import is still very slow and might need some more tuning).

DanielYang59:
image

Currently like 12% of the import time is used to import `import plotly.figure_factory` in `ptable.plotly` where it's only used once across the entire code base, I have lazily imported it ed48b1395e52b9969a17d20a4839e31431b2dda8.

After that:
image

The `sympy` import by `pymatgen` should also be done lazily https://github.com/materialsproject/pymatgen/pull/4128#issuecomment-2428964963. `scipy` doesn't seem to have much room for improvement because it's used a lots of place.

janosh:
> By the way, how do we regenerate the `.pytest-split-durations` record (not now, after we finish everything, current import is still very slow and might need more tuning).

the command for that is

```sh
pytest --store-durations --durations-path tests/.pytest-split-durations
```

i should have documented that somewhere. probably best in `test.yml` where we set `test-cmd: ...` as a YAML comment

DanielYang59:
Thanks a ton for letting me know! 

What I don't really understand is how do we generate this runtime record via GitHub CI runners (on contrast to run that command locally)?

Especially for `pymatgen` where:
- it's huge and I don't have that much resources to run all jobs locally
- tasks are split

I have a feeling that we may not want this test to run for each commit (because I don't expect we would introduce changes that change import time very often), it's slowing down everything. Perhaps we only run it only when merging into main?

For this PR, I currently cannot think of more improvement to make as the left is pretty much core packages like scipy, pandas, matplotlib. Do we want to perhaps merge it (after I modify it to run at main branch only)

janosh:
very thankful for this PR! 👍 didn't mean to keep it parked so long

> What I don't really understand is how do we generate this runtime record via GitHub CI runners (in contrast to run that command locally)?

i think generating locally should be fine. what matters shouldn't be absolute values for test run times but how long a test takes relative to the others which will be similar in CI and locally

> it's huge and I don't have that much resources to run all jobs locally

even pymatgen only takes 10 min or so to run the whole test suite locally for me. curious how long for you?

> Perhaps we only run it only when merging into main?

is the test slow enough in CI that it would extend overall time to checks complete if it ran inside its own split? i just tried `test_import_time` locally and got 80.52s. will be more on GitHub but tests already take ~2 min.
either way, i think it's fine to only run this test on `main`

is this PR ready to go from your side? are we waiting on a `monty` release to get the new startup benefits?

janosh:
> are we waiting on a `monty` release to get the new startup benefits?

ah, saw your comment in `pyproject`

```
    # TODO: pmv doesn't actually depend on monty, however latest monty
    # includes a critical import patch, remove this after pmg bump dep
    "monty>=2024.10.21",
```

DanielYang59:
> i think generating locally should be fine. 

> even pymatgen only takes 10 min or so to run the whole test suite locally for me. curious how long for you?

Yep should takes about the same time just like pymatgen CI ([macOS runner is running on M1 chip](https://docs.github.com/en/actions/using-github-hosted-runners/using-github-hosted-runners/about-github-hosted-runners), each split takes around 1 min), as I'm usually coding on a M3 MacBook Air laptop. I'm just being curious as I think running on a dedicated runner would give better scalability and consistency :)


> is the test slow enough in CI that it would extend overall time to checks complete if it ran inside its own split?

The original implementation was running 10 repeats, but I reduced repeat to 3 so should be good now.

tests/performance_tests/test_import_time.py
```diff
@@ -43,21 +43,18 @@
...
-        module_name: measure_import_time_in_ms(module_name)
+        module_name: round(measure_import_time(module_name), 2)
         for module_name in REF_IMPORT_TIME
     }
 
     # Print out the import times in a copyable format
     print("\nCopyable import time dictionary:")
-    print("{")
-    for module_name, import_time in import_times.items():
-        print(f'    "{module_name}": {import_time:.2f},')
```
DanielYang59: @janosh This part was intended to generate a directly copyable dict (for us to directly copy and paste to update the time recording), printing a dict directly seems to give a one-liner:

```python
Copyable import time dictionary:
{'pymatviz': 2084.25, 'pymatviz.coordination': 2342.41, 'pymatviz.cumulative': 2299.73, 'pymatviz.histogram': 2443.11, 'pymatviz.phonons': 2235.57, 'pymatviz.powerups': 2172.71, 'pymatviz.ptable': 2286.77, 'pymatviz.rainclouds': 2702.03, 'pymatviz.rdf': 2331.98, 'pymatviz.relevance': 2256.29, 'pymatviz.sankey': 2313.12, 'pymatviz.scatter': 2312.48, 'pymatviz.structure_viz': 2330.39, 'pymatviz.sunburst': 2395.04, 'pymatviz.uncertainty': 2317.87, 'pymatviz.xrd': 2242.09}
```

tests/performance_tests/test_import_time.py
```diff
@@ -69,12 +66,12 @@ def measure_import_time_in_ms(module_name: str, count: int = 3) -> float:
     """
     total_time = 0.0
 
-    for _ in range(count):
-        start_time = time.perf_counter_ns()
```
DanielYang59: using `perf_counter_ns` was intended to get more precious time measure, but admittedly for our case the error should be negligible

I personally prefer the original function name `measure_import_time_in_ms` as it's more obvious the time is given in ms not seconds.

tests/performance_tests/test_import_time.py
janosh: i saw that. the thing is, if you paste that printed dict into a python file, `ruff` will auto-format it to multiple lines so having easier to read code seemed more important

DanielYang59: Fair point!

PR #241, opened 2024-11-04
Improve missing value handling logic of `ptable_heatmap`

### LINKED ISSUE #230: Better missing value handling strategy for ptable heatmap
- Currently missing value would either be replaced by zero or mean, perhaps it's better to add an option to keep missing value as is (shown as `-`).
- User cannot control the missing value strategy from `ptable_heatmap`, which default to `mean`, leading to erroneous values and statistics.
https://github.com/janosh/pymatviz/blob/25f90bd0429526a7328c42e4a0e85b5dfc44646d/pymatviz/ptable/_process_data.py#L82

```python

import pymatviz as pmv
from pymatviz.enums import Key


df_expt_gap = load_dataset("matbench_expt_gap")
counted_data = pmv.count_elements(df_expt_gap[Key.composition], fill_value=None)

ax = pmv.ptable_heatmap(
    counted_data,
    value_show_mode="percent",
)
pmv.io.save_and_compress_svg(ax, "./ptable-heatmap.svg")
```
ptable-heatmap

Should be:
ptable-heatmap

DanielYang59:
## Improve missing value handling logic of `ptable_heatmap`

Our `ptable_heatmap` has suboptimal missing value handling logic, to close #230:
- Currently Non-`PTableData` typed data would be converted, but with the `missing_strategy` hard-coded as `mean`, meaning we're artificially creating non-existent data by default also user cannot control this behaviour, perhaps showing missing value as is (`-`) would be better.
- Is there a better way to resolve this than simply adding one more argument (`on_empty` sort of overlaps, perhaps we could partly reuse it)
- Import performance check show current import time, also round to int (should suffice as import time is on the `1000` level)

### TODO list:
- [ ] Make sure a warning would be emitted (only) when data contains `NaN/None`
- [ ] Test missing value handling for both `PTableData` and others (`pd.DataFrame`) when containing missing values (NaN/None)


janosh:
i'll go ahead and close this PR for now as we reached feature parity between the `plotly` and `matplotlib` `ptable` plots and i'm more inclined to maintain the `plotly` variants due to interactivity, HTML export and ease of testing

DanielYang59:
Thanks for letting me know, fully understandable and I might do back to this a bit later :) Good news is my TODO list for MP stack is mostly cleared :)

PR #248, opened 2024-11-19
[Breaking] Split `utils` into sub-modules, move `typing` from `utils` to root (`pmv.typing`)

DanielYang59:
Feel free to move things around.

- [x] Re-import everything from `__init__` to avoid breaking
- [x] Split tests as well
- [ ] Update the module docstring to include list of available func/...
- [x] Sort functions alphabetically
- [ ] Migrate examples (prefer more specific import? But still allow import directly from `pymatviz.utils`)


DanielYang59:
@janosh Perhaps let's finish this first, any comment on current categorizing? 😄

pymatviz/utils/testing.py
```diff
@@ -0,0 +1,8 @@
+"""Testing related utils."""
+
+from __future__ import annotations
+
+from pymatviz.utils import ROOT
+
+
+TEST_FILES: str = f"{ROOT}/tests/files"
```
DanielYang59, line 8: This might seem a bit overkill, but I kind of want to reserve this for more test functions like #208

pymatviz/utils/misc.py
janosh, line 1: i think `misc.py` is a bit too generic. no way of knowing what's in it without looking at the file. i think `ExperimentalWarning` can be safely moved into `utils/__init__.py` and this file deleted

pymatviz/utils/image.py
janosh, line 1: i would move `luminance` and `pick_bw_for_contrast` these functions into `utils/plotting.py` and remove this file

assets/scripts/coordination/coordination_hist.py
```diff
@@ -8,7 +8,7 @@
 
 import pymatviz as pmv
 from pymatviz.coordination import CnSplitMode
-from pymatviz.utils import TEST_FILES
+from pymatviz.utils.testing import TEST_FILES
```
janosh, line 11: maybe reexport `TEST_FILES` from `utils/__init__.py`

janosh (commented):
the changes here look great! just some minor suggestions

janosh (approved):
the changes here look great! just some minor suggestions

assets/scripts/coordination/coordination_hist.py
DanielYang59: I kind of want to hide test utils like this from the "public" interface (`pmv.utils`), what do you think?

pymatviz/utils/misc.py
janosh: > hold anything leftover that don't have a category of it own

i think having code that doesn't fit a category is often a bad sign. means you may need to rethink the structure of a code base. so i'd rather not have this file and force meaningful categorization

pymatviz/utils/image.py
DanielYang59: I was trying to do a more fine-grained splitting for `plotting` if possible because this PR seems to split the `utils` into basically `data/plotting` and a small part into `typing`, it's an improvement but still the two modules are pretty big. Also I cannot split `plotting` into `mpl/plotly` because most functions support both. 

Perhaps just take this first step here and do more in the future when needed?

pymatviz/utils/misc.py
DanielYang59: Okay that's the fair point

assets/scripts/coordination/coordination_hist.py
janosh: sounds good. i moved `pymatviz/utils/typing.py` to `pymatviz/typing.py` since i think types are important enough to not be relegated into `utils`. we also now match structure with `numpy.typing`

DanielYang59: That's a good idea, but would be breaking?

janosh: good point! still, i think we should go ahead. i imagine very few people import `pymatviz` types in external code bases anyway.

DanielYang59:
Only one task left, I would add a list of available function to `utils.plotting` tomorrow

DanielYang59:
Thanks! I could sleep with my eyes close now :)

janosh:
thanks so much for your help with this! 🙏

DanielYang59:
No problem, enjoy your day (and my sleep), thanks for reviewing and all the comments

DanielYang59:
Looks like the fix for `GitHub Pages / build` is working now (one extra `/`), but import performance test is failing for some reason, I would have a closer look tomorrow. Also it seems I forgot to display current time (`pymatviz import too slow! hard_threshold=3126.38 ms`), only the expected time is shown

janosh:
i added the expected time. [it now prints](https://github.com/janosh/pymatviz/actions/runs/12051121078/job/33601461364)

> FAILED tests/performance_tests/test_import_time.py::test_import_time - Failed: pymatviz import too slow! took 3722 ms, hard_threshold=3126 ms

so at least it's not a massive regression. i'm guessing it's because i added some new modules in `pymatviz.classify` in #252 and #253 which now need to be parsed so perhaps the slow down is expected? still would be good to find out exactly. could also be i'm importing some module at the top-level which before was only imported at run time. maybe `import sklearn.metrics as skm`. if that's the case then the test did it's job perfectly :)

DanielYang59:
> i added the expected time. [it now prints](https://github.com/janosh/pymatviz/actions/runs/12051121078/job/33601461364)

Thanks for adding this, I also added one in #241.

> 'm guessing it's because i added some new modules in pymatviz.classify in https://github.com/janosh/pymatviz/pull/252 and https://github.com/janosh/pymatviz/pull/253 which now need to be parsed so perhaps the slow down is expected?

I just looked into the commit history and it appears I updated the import time in a6302f2a17670079ca9101ec8847f03661c7d808 (**I'm completely confused because I cannot remember I re-ran import time benchmark in this PR at all**)

I don't think this should be updated because I don't expect import time decrease, by the time the import time tester was added, the `monty` fix had been applied. Also this cannot be "accidentally" updated because the import time test would print a copiable dict instead of overwriting in place.

In a word, I'm absolutely confused why I updated the import time baseline and where that data came from, perhaps let's revert it?

janosh:
thanks for checking! i reverted to the old timings in bc1bc29

DanielYang59:
Thanks for the revert, though I'm still super confused where the data came from ...

PR #251, opened 2024-11-22
Add `brillouin_zone_3d` plot function

janosh:
See [`pymatviz/brillouin.py`](pymatviz/brillouin.py).

| [`brillouin_zone_3d(struct)`](assets/scripts/brillouin/brillouin_zone_3d.py) | [`brillouin_zone_3d(struct, system="hexagonal")`](assets/scripts/brillouin/brillouin_zone_3d.py) |
| :--------------------------------------------------------------------------: | :----------------------------------------------------------------------------------------------: |
|                         ![brillouin-cubic-mp-10018]                          |                                 ![brillouin-hexagonal-mp-862690]                                 |
|                      ![brillouin-monoclinic-mp-1183089]                      |                               ![brillouin-orthorhombic-mp-1183085]                               |

[brillouin-cubic-mp-10018]: https://github.com/user-attachments/assets/d08d3984-7c97-4eb0-a466-e957f21a7291
[brillouin-monoclinic-mp-1183089]: https://github.com/user-attachments/assets/1cb58244-2541-4b7c-bc61-8251336eb6be
[brillouin-hexagonal-mp-862690]: https://github.com/user-attachments/assets/026ba388-e6d0-4195-b2fe-885f7d265d43
[brillouin-orthorhombic-mp-1183085]: https://github.com/user-attachments/assets/079078c6-f972-48de-9da9-84c29fb85597


DanielYang59:
Thanks for asking me, this one looks gorgeous (both the code and figures) and surely would be super helpful for visualizing Brillouin zones. I just did a quick search and it seems [one package provides some functionality](https://github.com/chstan/arpes/blob/master/docs/source/brillouin-zones.rst) but only limited crystal systems, so we might be confident that we're not recreating wheels.

Regarding the API, I personally seldom plot on a given `fig` (I prefer to plot individual figure/subplots on a separate canvas for stability) so I might prefer taking `fig` and `subplot_idx` as optional keyword args (instead of positional). Apart from this everything looks good and clean to me (one `kwargs` for one "element").

I guess you have caught the appearance issues like missing arrow, basis vectors flying out of the canvas, random white lines cutting the edges (wrong trace order I guess?) and perhaps the edges at the back should be dotted?

janosh:
> I just did a quick search and it seems [one package provides some functionality](https://github.com/chstan/arpes/blob/master/docs/source/brillouin-zones.rst) but only limited crystal systems (I cannot install this package with pip)

thanks for checking. i also looked around before starting the implementation but didn't find `arpes`. the ones i saw were

- https://github.com/ots22/brillouin (only 2D)
- https://github.com/hedhyw/BrillouinZones (only `matplotlib`)

neither had an interactive (drag and zoomable) 3D version implemented with `plotly` which is what i need.

> having two falsy values might be a bit confusing, would using Literal["off"] | None be more self-descriptive?

yeah, need to think about that

> so I might prefer taking fig and subplot_idx as optional keyword args (instead of positional)

good point. that was hastily done to make the example figure above but i was also considering removing the fig arg entirely. the reason it might make sense to keep it as optional arg is that one common use case is to plot the BZ into a (phonon or electronic) bandstructure plot to visualize the path through k-spce being sampled by the bands. in which case this function might frequently be used to make small overlayed subfigures.


> and perhaps the edges at the back should be dotted?

that would be cool! but i think it's quite hard since you can rotate the figure, meaning the line style needs to change from solid to dotted dynamically

DanielYang59:
> thanks for checking. i also looked around before starting the implementation but didn't find arpes. the ones i saw were

Yes I found these two as well, I guess there's no existing package for this.

> that would be cool! but i think it's quite hard since you can rotate the figure, meaning the line style needs to change from solid to dotted dynamically

Always easier said (me) than done. I never touched `plotly` before but it looks good, I would have a try on it as well :)

janosh:
> Always easier said (me) than done. I never touched plotly before but it looks good, I would have a try on it as well :)

definitely worth trying. the interactivity is super helpful. and the fact that you can print to PDF and export as HTML for websites makes it very versatile. plus the API is better designed and less confusing than that of `matplotlib`. definitely worth learning

DanielYang59:
This is truly amazing quality plot, thanks so much for the work :)

janosh:
don't have time right now but next step is to add a `brillouin_zone: dict | False` keyword to the `phonon_bands` plot that allows overlaying or plotting next to the bands the Brillouin zone to show what path in reciprocal space the bands correspond to. a bit [like this one](https://www.tf.uni-kiel.de/matwis/amat/semi_en/kap_2/backbone/r2_1_5.html):

Screenshot 2024-11-30 at 6 46 09 AM

PR #263, opened 2025-01-24
Remove hard-coded `gridsize` in `density_hexbin`

jcwang587:
custom `gridsize` example:

```py
import pymatviz as pmv

dummy_data = pmv.data.regression()
pmv.utils.apply_matplotlib_template()

ax = pmv.density_hexbin(
    dummy_data.y_pred,
    dummy_data.y_true,
    best_fit_line={"annotate_params": {"loc": "lower center"}},
    gridsize=40,
)
```
density-scatter-hex


janosh (approved):
thanks @jcwang587! 👍

PR #271, opened 2025-02-24
fix error in docstring

GKarbon:
Fixed incorrect docstring in `pymatviz.phonons.phonon_bands` where the default value of `path_mode` parameter was incorrectly documented as `'union'` when it should be `'strict'`.


janosh (approved):
good catch, thanks @GKarbon 👍

PR #293, opened 2025-05-18
Add `cell_size_fn: Callable[[ModuleStats], float]` keyword to `py_pkg_treemap`

janosh:
allows scaling cell size by quantities other than line count (such as number of functions, classes, methods, imports, etc.). maybe of interest to @mkhorton

py-pkg-treemap-pymatgen

```py
import pymatviz as pmv

pmv.set_plotly_template("plotly_white")

package = "pymatgen"
fig = pmv.py_pkg_treemap(package)
fig.layout.title.update(text=f"{package} Package Structure", font_size=20, x=0.5)
fig.show()
```


DanielYang59:
thanks for pinging me, this figure looks great (I haven't looked into the source code yet)!

One small issue though: when the tile is scaled, the font is scaled too which makes the fonts in small tiles hardly readable

Not sure if we to allow manually pass a "weight" to scale certain module though, for example we might determine certain module (core for example) is very important but doesn't have many lines of code, but we do want to make it bigger?

janosh:
> (core for example) is very important but doesn't have many lines of code, but we do want to make it bigger?

you could do this now with the `cell_size_fn`

```py
cell_size_fn=lambda cell: cell.line_count * (2 if "core" in cell.full_module else 1)
```

but it would break the representativeness of cell sizes. the goal of the plot is to show where most of the code/functionality is. the importance of that functionality is probably better communicated by the name (e.g. `core` in this case is pretty clear about its importance)

> One small issue though: when the tile is scaled, the font is scaled too which makes the fonts in small tiles hardly readable

you can pass `min_lines` to exclude files below a certain threshold. should probably rename `min_lines` to `min_size` now since it's no longer always the line count that determines cell size. this is what `min_lines=400` looks like:

py-pkg-treemap-pymatgen

PR #298, opened 2025-05-24
Add element color legend to plotly 2d and 3d structure visualizations

janosh:


DanielYang59:
Hi @janosh thanks for pinging me

> whether the color legend is an improvement over element symbol annotations on every site

Wow the color legend is definitely an improvement to me, it's much cleaner (avoided symbol duplicate and overlapping)

janosh:
thanks, i'm thinking the output might be decent enough now to warrant adding proper mime-type handling to make Jupyter/iPython aware that returning a structure from a notebook cell should render it using `structure_3d_plotly`. we could add the similar associations for `pymatgen.phonon.bandstructure.PhononBandStructureSymmLine`->`phonon_bands`, `pymatgen.phonon.dos.PhononDos`->`phonon_dos`, etc. for XRD, ASE Atoms, etc.

Issue #307, opened 2025-07-05
give validators error


Asif-Iqbal-Bhatti:
Hello Janosh,

When i run the heatmap module, it gives me this error

'''
from plotly.validators.scatter.marker import SymbolValidator
ModuleNotFoundError: No module named 'plotly.validators.scatter'
'''

I have installed the latest version of pymatviz and plotly. In the latest version of Plotly, I see there is no plotly.validators.scatter.marker. I cannot locate where it moves. Can you check? If this is right, I am the only one who is facing this error.

Also, when I use this importing
from pymatviz import count_elements, ptable_heatmap, ptable_heatmap_plotly

I don't see a count_elements or ptable_heatmap folder. Where are these attributes are located? 

Regards,
Asif

PR #310, opened 2025-07-09
Fix plotly validators

bmaranville:
In `plotly==6.2`, the directories of validator modules were removed in favor of a system that generates validator instances on demand from `plotly.validator_cache.ValidatorCache`.  The validator cache was already used internally in plotly > 99% of the time, and now it is the only way to get a validator.

In this PR, the uses of directly imported validator classes is refactored to use the `ValidatorCache`.  This is backward-compatible with `plotly<6.2`, because the `ValidatorCache` was already in place for e.g. `plotly==6.1.2`

## Summary by CodeRabbit

* **Chores**
  * Improved efficiency by centralizing the handling of Plotly validators with a caching approach.
  * Updated dependency requirements to support all Plotly versions 6 and above, including version 6.2.

* **Tests**
  * Aligned test imports with the new validator caching method.

### DIFF
--- a/pymatviz/cluster/composition/plot.py
+++ b/pymatviz/cluster/composition/plot.py
@@ -17,6 +17,7 @@ from pymatgen.core import Composition
 from pymatviz.cluster.composition.embed import matminer_featurize, one_hot_encode
 from pymatviz.cluster.composition.project import project_vectors
 
+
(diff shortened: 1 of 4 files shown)


bmaranville:
This addresses #307

janosh (approved):
wonderful! thanks so much for this PR 👍

